PR #2, opened 2019-12-11
Proposal for a behavioral change of the testing service

johannes-b:


text/0002-test-finished-with-test-result.md
```diff
@@ -0,0 +1,53 @@
...
+  "id": "f2b878d3-03c0-4e8f-bc3f-454bc1b3d79d",
+  "time": "2019-06-07T07:02:15.64489Z",
+  "contenttype": "application/json",
+  "shkeptncontext": "08735340-6f9e-4b32-97ff-3b6c292bc509"
+}
+```
+
+## Internal details
+
+* *How the change would impact and interact with existing functionality?*: Since the jmeter-service does not send the `sh.keptn.events.evaluation-done`, the lighthouse-service is responsible for this step. To make sure, the lighthouse-service could receive the test result and skips the process of querying the SLI-provider since no monitoring data from the test would be available.
```
christian-kreuzberger-dtx: Actually, one could still want to query the SLI-provider for data, as monitoring data for the test could be available.

johannes-b: Valid point! 
This is the reason why I used 'could' in the last sentence.

christian-kreuzberger-dtx: Then I would like to ask you to move that sentence into the **Future possibilities** section (and to re-phrase it such that it can stand on its own).

Leave the part with 
> Since the jmeter-service does not send the `sh.keptn.events.evaluation-done`, the lighthouse-service is responsible for this step. 

as it is.

johannes-b: Okay. I moved the sentence to the *Future possibilities* section.

christian-kreuzberger-dtx:
Merging this. @johannes-b can you please also create a PR for the spec and link to this KEP.

Thanks!

PR #4, opened 2020-01-28
KEP 04 - Enrich SLI with display name and deep link

johannes-b:


text/0004-enriched-indicator-value.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Explanation
+
+Instead of limiting an indicator to a single value, this KEP proposes to enrich the event payload of [sh.keptn.internal.event.get-sli.done](https://github.com/keptn/spec/blob/master/cloudevents.md#get-sli-done) with a set of values and a deep link for each indicator, as shown below:
+
+```yaml
+    "indicatorValues": [
+      {
+        "metric":"request_latency_p95",
+        "values": [1.162, 1.651, 1.162, 1.653, 1.192, 1.191],
```
christian-kreuzberger-dtx: Where did `value` go? I believe we should have both, `value` as a required field, and `values` should be optional.

In addition, just `values` would be the `y` axis, I think we would also need an `x` axis for those values, or at least define  that they are evenly spread between `start` and `end` datetimes of the get-sli event.

text/0004-enriched-indicator-value.md
```diff
@@ -0,0 +1,63 @@
...
+        "origin":"https://xyz.com/#smgd;sci=SERVICE-18E3F3E463387767;tab=RT;servicefilter=0%1E10%11SERVICE_METHOD_GROUP-175CF39880103983;gf=all;gtf=l_24_HOURS;timeframe=custom1580161200000to1580168400000"
```
christian-kreuzberger-dtx: I wonder if `origin` is a good choice. Thinking about future use cases, we might want to have values come from multiple data sources, then `origin` would have a different meaning. How about `details`, or, as it is an url, even `detail_url`?

text/0004-enriched-indicator-value.md
```diff
@@ -0,0 +1,63 @@
...
+
+* *How the change would impact and interact with existing functionality?*: The lighthouse-service, which is responsible for processing SLIs, has to be capable of dealing with the set of values. Besides, the Keptn's bridge can be enhanced by showing the deep link next to the visualization of an SLI. 
+
+## Trade-offs and mitigations
+
+*What are some drawbacks? What are some ways that they might be mitigated?*: A drawback is the amount of data stored by Keptn. The internal data store of Keptn should not grow due to data that is not needed. There should be an agreement on how many values for an indicator are provided (see open question).
+
+## Breaking changes
+
+There is no breaking change since this KEP adds two properties to an event, but does not remove or rename a property.
```
christian-kreuzberger-dtx: If you add the `value` property back to the payload, then there is no breaking change.

text/0004-enriched-indicator-value.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Open questions
+
+Sending all data that is available for an indicator does not make sense since it is then a duplicate to the original data set. The origin of the data has to be the monitoring/testing tool that provides the data for the SLI-provider. It should be clarified, how many values make sense.
```
christian-kreuzberger-dtx: To be honest, almost any number of values makes sense depending on the use-case. We (or much more the end-user that uses Keptn) will have to live with the duplication if they want this feature.
I would however suggest to make it configurable whether these values should be fetched from the SLI provider or not.

Another open question for me would be:

**How long to store the results?**

johannes-b:
An issue in an SLI-provider also references this idea of enriching SLIs: https://github.com/keptn-contrib/dynatrace-sli-service/issues/35

johannes-b:
No progress on that enhancement proposal. I closed it.

PR #5, opened 2020-02-24
KEP for Deployment Finished Event Extension with Deployment Endpoint URL

grabnerandi:
See my proposal for this KEP 0005!

### DIFF
--- a/text/0005-deployment-finished-with-deployed-endpointurl.md
+++ b/text/0005-deployment-finished-with-deployed-endpointurl.md
@@ -15,9 +15,10 @@ As these tools are currently sending the Deployment Finished Event to Keptn I pr
 
 ## Internal details
 
-I propose that we extend the Deployment Finished Event with a new field called "deployedEndpoint" (happy to have a discussion on how this field should be named in the end). Deployment Tools that send a deployment finished event to Keptn SHOULD (but are not obliged) to fill this field.
-If this field IS PROVIDED keptn services such as the JMeter, Neotys ... service can use this URL to execute their tests against.
-If this field IS NOT provided then services such as JMeter, Neotys ... can fall back to their current "guestimation" of the URL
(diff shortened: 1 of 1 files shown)


text/0005-deployment-finished-with-deployed-endpointurl.md
```diff
@@ -0,0 +1,41 @@
...
+No breaking changes as we just add a new optional field.
+
+## Prior art and alternatives
+
+Not aware of any alternatives other than our current "guestimage" approach
+
+## Open questions
+
+How shall we name that field?
+Is one field enough or shall we provide a set of fields, e.g: root endpoint, health check endpoint ...?
```
christian-kreuzberger-dtx: I think one field with the base url is enough, anything else is subject to configuration of the testing service (e.g., jmeter-service).

johannes-b: I agree with using only one field and like the already proposed property name: `deployedEndpoint`.

johannes-b:
I like the proposed KEP, as it helps to separate concerns. (i.e., it is not the test-service's responsibility to come up with the endpoint to test).

text/0005-deployment-finished-with-deployed-endpointurl.md
christian-kreuzberger-dtx: How about `deploymentURL`, or to be even more generic (it does not need to be a URL): `deploymentURI`?

grabnerandi: deploymentURI works for me!

johannes-b: The team agreed on introducing two properties: `deploymentURILocal` and `deploymentURIPublic` (to allow tests on dark deployments). Please read the update in the section *Internal details*.

PR #13, opened 2020-03-18
KEP 13 - Provide an evaluation result history API call

grabnerandi:


text/0006-get-evaluation-history-api-call.md
```diff
@@ -0,0 +1,61 @@
...
+The Keptns bridge is doing a great job in visualizing historical evaluation results - but - it is doing it by fetching data directly from the MongoDB. We should build this logic into an official API that the bridge is using and that others can use as well, e.g: users that want to visualize the data in a similar way in other tools such as Jenkins, Azure DevOps ...
+
+## Details
+
+Here some ideas on how that API Call could look like!
+
+Examples:
+Default: Last 20 runs
+```
+/project/sockshop/stage/dev/service/carts/evaluations
```
christian-kreuzberger-dtx, line 17: Hi,

I believe this endpoint already existed back in Keptn 0.6.x (it definetely was there in 0.7.x). In 0.8.x a query looks like this:
```
GET /api/mongodb-datastore/event/type/sh.keptn.event.evaluation.finished?filter=data.project:sockshop AND data.service:carts AND data.stage:staging&excludeInvalidated=true&limit=50
```

There is also a parameter called `fromTime` (though no `to` param).
Anyway, I'm not sure how sophisticated the filtering is (e.g., using `from` and `to` or even `labels`), but I guess that those filters are depending on MongoDB queries, so quite a bit might be possible.

Could you please check out how many of your requests are handled by this endpoint?

christian-kreuzberger-dtx:
Closing this PR, as it should already be implemented in mongodb-datastore.

In 0.8.0 a query looks like this:
```
GET /api/mongodb-datastore/event/type/sh.keptn.event.evaluation.finished?filter=data.project:sockshop AND data.service:carts AND data.stage:staging&excludeInvalidated=true&limit=50
```

There is also a parameter called `fromTime` (though no `to` param).

PR #16, opened 2020-04-07
KEP 16 - Add default SLO and SLI files to a service

johannes-b:
## Motivation

When starting with Keptn SLIs & SLOs are a bit like "magic". By default there is no visible SLI and therefore it is hard to understand what is happening behind the scenes and it is also hard to get started with a custom SLI because there is no good template to start from.
The motivation therefore is to make it easier for users to understand and extend SLIs & SLOs by automatically adding a default sli & slo when onboarding a new service.

Please read the KEP for more information.


johannes-b:
I just want to point to an issue: https://github.com/keptn/keptn/issues/1257 which is already in the dev pipe and planned for 0.7.0. The referenced issue is half of the story of your proposal. 

I agree with creating a default SLO for a new service.

PR #18, opened 2020-06-18
KEP 18 - Simplify installer - do not install istio nor nginx-ingress

christian-kreuzberger-dtx:
This Draft provides some information on why we should simplify the Keptn installer to no longer install istio nor nginx-ingress.

FYI, I believe that implementing this KEP will make the following Keptn Issues obsolete:

* https://github.com/keptn/keptn/issues/1820
* https://github.com/keptn/keptn/issues/1003
* https://github.com/keptn/keptn/issues/568
* https://github.com/keptn/keptn/issues/1406
* https://github.com/keptn/keptn/issues/1825

*Please consider the open questions on the bottom of the markdown file*

### DIFF
--- a/text/0018-simplify-installer.md
+++ b/text/0018-simplify-installer.md
@@ -139,7 +139,7 @@ If the user wants to change this behaviour afterwards, it is as easy as calling 
     kubectl patch svc bridge -n keptn -p '{"spec": {"type": "NodePort"}}'
     ```
 
-If the user wants to use an ingress-controller (e.g., nginx-ingress, Linkerd, istio, traefic) for API/Bridge, the recommended way is to use ClusterIP and apply an ingress manifest manually. 
+If the user wants to use an ingress-controller (e.g., nginx-ingress, istio, traefik) for API/Bridge, the recommended way is to use ClusterIP and apply an ingress manifest manually. 
 
 
 ## Breaking changes
@@ -164,6 +164,8 @@ Please note: This is not a tutorial, just a sketch of what it looks like to the 
 
 Keep in mind that installing a production setup for Istio is not trivial, and the instructions here are only meant for **demo purposes**.
 
+Also, SSL/TLS support is no longer part of a Keptn installation.
+
 ### Full Installation on GKE - Variant 1: Use Google LoadBalancer
 
 1. Ensure that you are connected to the correct Kubernetes cluster using `kubectl get nodes`.
@@ -173,6 +175,26 @@ Keep in mind that installing a production setup for Istio is not trivial, and th
    ```console
    istioctl install --set profile=demo
    ```
+1. [Not sure where to put this] Create a public-gateway for istio
+   ```yaml
+   ---
+   apiVersion: networking.istio.io/v1alpha3
+   kind: Gateway
+   metadata:
+   name: public-gateway
+   namespace: istio-system
+   spec:
+   selector:
+      istio: ingressgateway
+   servers:
+   - port:
+         name: http
+         number: 80
+         protocol: HTTP
+      hosts:
+      - '*'
+   ```
+   SSL/TLS is also possible, but not part of the demo.
 1. Install Keptn and automatically expose the API using LoadBalancer (recommended for GKE)
    ```console
    keptn install --keptn-api-service-type=LoadBalancer # --use-case=delivery will most likely be needed for 0.7.x
@@ -196,7 +218,7 @@ Keep in mind that installing a production setup for Istio is not trivial, and th
    ```
    Note down IP and Port (e.g., 5.6.7.8, port 80) and set a hostname for it in `INGRESS_HOSTNAME_SUFFIX` in the configmap
    ```console
-   kubectl -n keptn create configmap --from-literal=INGRESS_HOSTNAME_SUFFIX=5-6-7-8.nip.io
+   kubectl -n keptn create configmap ingress_hostname --from-literal=INGRESS_HOSTNAME_SUFFIX=5-6-7-8.nip.io
    ```
 1. Follow the tutorial for creating a project and onboarding a service
 
@@ -212,6 +234,26 @@ This kind-of rebuilds what keptn 0.6.x will do out of the box.
    ```console
    istioctl install --set profile=demo
    ```
+1. [Not sure where to put this] Create a public-gateway for istio
+   ```yaml
+   ---
+   apiVersion: networking.istio.io/v1alpha3
+   kind: Gateway
+   metadata:
+   name: public-gateway
+   namespace: istio-system
+   spec:
+   selector:
+      istio: ingressgateway
+   servers:
+   - port:
+         name: http
+         number: 80
+         protocol: HTTP
+      hosts:
+      - '*'
+   ```
+   SSL/TLS is also possible, but not part of the demo.
 1. Install Keptn, but do not expose the API:
    ```console
    keptn install --keptn-api-service-type=ClusterIP # --use-case=delivery will most likely be needed for 0.7.x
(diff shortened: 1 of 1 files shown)


jetzlstorfer:
for convenience, here is the link to the formatted doc: https://github.com/keptn/enhancement-proposals/blob/1a6f3e2b3f4d4dc697c12a622e223c4862fd7afc/text/0018-simplify-installer.md

text/0018-simplify-installer.md
```diff
@@ -0,0 +1,152 @@
...
+* CLI: After installation, try to access the API via all possible IP addresses that the installer determined. If nothing works, show the user a warning and a link to the troubleshooting guide in our docs.
+* CLI: Installation should be considered successful if all deployments/pods have been started. Being able to reach Keptn API is not a strong requirement, and should only print a warning.
+* Installer: Remove any manifests and any scripts that make use of Istio or nginx-ingress
+* Installer: use a placeholder for the service-type of `api-gateway-nginx` and `bridge`, which is then applied based on the settings `--keptn-api-service-type` and `--keptn-bridge-service-type`
+* Installer: Determine all possible IP addresses (and ports) that the user can use to access Keptn API (internal IP, external IP, ...)
+* Installer: Do not set `KEPTN_DOMAIN`
+* Installer: Do not generate SSL certificates
+
+* CLI/API: Change `keptn configure bridge` no longer provide the flag `action=[expose | lockdown]`, but a flag to configure the service-type [user and password should stay as they are, they are still required]
```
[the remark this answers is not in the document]
christian-kreuzberger-dtx: > I would propose that neither the CLI nor the API can change the service type because this should be handled by an administrator.

I agree with that proposal. We should document the command, and remove the funcionality from our API.

text/0018-simplify-installer.md
```diff
@@ -0,0 +1,152 @@
...
+
+This would be the default when just using `keptn install` (or `keptn install --keptn-api-service-type=ClusterIP --keptn-bridge-service-type=ClusterIP`):
+
+```
+$ kubectl -n keptn get service bridge api-gateway-nginx
+NAME                TYPE        CLUSTER-IP    EXTERNAL-IP   PORT(S)          AGE
+bridge              ClusterIP   10.0.83.209   <none>        8080:32122/TCP   85d
+api-gateway-nginx   ClusterIP   10.0.94.251   <none>        80/TCP           41d
+```
+In this example neither bridge nor API are reachable from the outside. To reach it, one would have to
```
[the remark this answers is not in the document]
christian-kreuzberger-dtx: Added.

text/0018-simplify-installer.md
```diff
@@ -0,0 +1,152 @@
...
+* General: `KEPTN_DOMAIN` is no longer available (the installer can not determine this reliably). We need to research where this might be a problem. For `helm-service` and `openshift-route-service`, we could use a separate configmap with a variable called `INGRESS_HOSTNAME_SUFFIX` which is set to `cluster.svc.local` by default.
+* Dynatrace-Service: `KEPTN_DOMAIN` is required for configuring problem notifications that are sent back to the Keptn API. We need to come up with a solution for this...
+
+* Docs: Update docs for installation with the preferred service-type for each Cloud Provider 
+* Docs: Remove the experimental tutorial about reusing an existing ingress-gateway (e.g., istio)
+* Docs/Tutorials: Update docs for installation with a troubleshooting guide if API can not be reached (e.g., refer to Kubernetes services and description of ClusterIP vs. NodePort vs. LoadBalancer) 
+* Docs/Tutorials: Provide solid Docs for Demo/POC (e.g., with GKE, K3s) and setting up a full installation of Keptn with onboarding services 
+
+
+From a technical perspective, how do you propose accomplishing the proposal?
```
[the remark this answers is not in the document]
christian-kreuzberger-dtx: Oops, done!

text/0018-simplify-installer.md
```diff
@@ -0,0 +1,172 @@
...
+If the user wants to use an ingress-controller (e.g., nginx-ingress, Linkerd, istio, traefic) for API/Bridge, the recommended way is to use ClusterIP and apply an ingress manifest manually. 
+
+
+## Breaking changes
+
+* The CLI command `keptn configure bridge` will need to be simplified to no longer change any service types or ingress manifests.
+* The configmap for `KEPTN_DOMAIN` will no longer be available, as we don't know the domain when installing Keptn.
+* Helm-Service will break without `KEPTN_DOMAIN` - instead we should ask the user to supply this configuration using a new environment variable called `INGRESS_HOSTNAME_SUFFIX` which we will default to `svc.cluster.local`
+* Helm-Service will need to check if istio is available before applying any virtualservices
+* Not sure, but OpenShift support might break and we need to fix certain things
```
bacherfl, line 151: Exposing the API and bridge should become easier in that case, since we can refer to the `oc create route` command. When onboarding a service, we also need the `INGRESS_HOSTNAME_SUFFIX'  because the openshift-route service will create the routes for each onboarded service, and those all refer to the istio-ingressgateway

text/0018-simplify-installer.md
```diff
@@ -0,0 +1,172 @@
...
+* Dynatrace-service will break as it also relies on `KEPTN_DOMAIN` - especially for problem notifications
```
bacherfl, line 152: In addition to Problem notifications, the `KEPTN_COMAIN` is currently used during the dashboard setup (to provide links to the onboarded services, and to the bridge), as well as to provide deep links to the bridge in dynatrace events.

For the first two cases, we could pass the URL of the API/Bridge as parameter(s) for the `configure monitoring` command.
For adding the deep links to the bridge in deployment events, we will have to rely on the `INGRESS_HOSTNAME_SUFFIX` env var (as we also do in the helm-service)

christian-kreuzberger-dtx: Thanks for bringing this up, I've added it to the KEP.
Basically, with the proposed change in this KEP deep links to Keptn Bridge will not work anymore, as Keptn does not expose Bridge via a VirtualService or alike.

text/0018-simplify-installer.md
christian-kreuzberger-dtx: Thanks, added.

christian-kreuzberger-dtx:
Thanks Andi, I'll add a section that describes what the installation feels like for the user in terms of a full installation.

text/0018-simplify-installer.md
christian-kreuzberger-dtx: I've also added instructions using `oc create route` for exposing api

christian-kreuzberger-dtx:
I've added instructions for GKE, OpenShift and K3s.

checkelmann:
Hi,

I understand the need to simplify the installer and that not every flavor of different ingress and virtual service types could be covered.

As @christian-kreuzberger-dtx also mentioned within the KEP is that different cloud providers got different types of LoadBalancer Annotations. This is the same for ingresses.

Perhaps it would be an idea instead of creating the ingress or virtual service to print out a default ingress yaml with preconfigured services hosts and ports, which could be then adjusted by the administrator or by a CI/CD Tool when installing keptn automatically.

Something like

```
keptn describe ingress
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
  name: kubernetes-api-ingress
spec:
  rules:
  - host: api.kubernetes.keptn.dev.ert.com
    http:
      paths:
      - backend:
          serviceName: kubernetes
          servicePort: 443
        path: /
  tls:
  - hosts:
    - api.kubernetes.keptn.dev.ert.com
```

christian-kreuzberger-dtx:
> Perhaps it would be an idea instead of creating the ingress or virtual service to print out a default ingress yaml with preconfigured services hosts and ports, which could be then adjusted by the administrator or by a CI/CD Tool when installing keptn automatically.

We will be trying to cover this with tutorials and docs with examples for now, which we believe is more than sufficient. Users should know their way around Kubernetes services and ingress if they want to use it. In fact, we have plenty of users that know what they are doing and overwrite the ingress gateway anyway, and we have plenty of users that struggle in determining the correct hostname for the ingress gateway on their Kubernetes setup.

PR #22, opened 2020-07-16
KEP 22 - Declarative configuration of Keptn

thschue:


thschue:
obsolete - successor is KEP65

PR #35, opened 2021-02-17
updated sequence triggers in next gen shipyard kep

warber:
This PR updates the next gen shipyard kep to use the agreed format for sequence triggers.

Signed-off-by: warber <bernd.warmuth@dynatrace.com>


johannes-b (approved):
lgtm

PR #37, opened 2021-03-18
KEP 37 - Closed loop remediation with Shipyard v0.2

johannes-b:
# Closed loop remediation with Shipyard v0.2

**Success Criteria:** A remediation process is modeled in the Shipyard of a project

## Motivation

* *Target*: As a user, I can model my custom remediation process in the Shipyard. The rem. process follows the closed-loop principle; meaning that it loops as long as the problem is not fixed and actions are available. 

* *Pain*: Currently, our opinionated closed-loop rem. process is hard-coded in the remediation-service. This limits a user in tailoring it. + It is not possible (!) to register action-providers to action.triggered events since they are not managed by the shipyard-controller.

* *Driver*: The use-case of automating remediations should work like the delivery (quality-gates) use-case.

## User Stories

* As a user, can define a remediation process (sequence) in the Shipyard file.
  - Depending on the number of actions defined in the remediation.yaml, this sequence gets triggered except for a passed evaluation (that considers closed problems)
* As a user (who is running an action-provider on the exec. plane), I can subscribe my action-provider to an `action.triggered` event. 
  - See issue: https://github.com/keptn/keptn/issues/3498
* *Changed behavior* As a user, I would like to define the mapping between problem and remedation action based on the root cause (spec change: add ProblemDetailsJSON) and not the problem title. (Consider updates of examples / tutorials)

## Remediation in Shipyard v0.2

```
- name: "production" 
  sequences: 
  - name: remediation
      triggeredOn: 
      - event: production.remediation.finished
          selector:
            match:
              evaluation.result: fail
      tasks:
    - name: get-action 
    - name: action
    - name: evaluation
      triggeredAfter: "10m"
      properties:
        timeframe: "5m"
```

## Open Question

- How can we stop the loop when, for example, the action returned `result=fail`?


johannes-b:
**Breakdown of Value Increment into Dev Epics (Issues):**

- Implement a `wait-service` that subscribes to the `wait` task
- Adapt the dynatrace-service to consider the ProblemDetailsJSON property containing the problem root cause
- Clean-up remediation-service to not control the workflow: just get the correct remediation action based on the problem root cause and current remediation progress
- Prepare an SLO/SLI file that checks whether the Problem has been closed (as part of the remediation process)
- Are there updates of the current action-providers (helm-service / unleash-service) necessary? 
- Update of tutorials/examples/docs/spec to support the remediation process modeled via Shipyard v.0.2

johannes-b:
## keptn/keptn: Implement a `wait-service` that subscribes to the wait task 

**Context:** A remediation sequence triggers an action that should resolve the problem. However, the effect of the action may not be visible immediately. Consequently, it is required to wait for a few minutes before triggering the next task of a rem. sequence. 

*Example:*

```
- name: "production" 
  sequences: 
  - name: remediation
      triggeredOn: 
      - event: production.problem.opened
      tasks:
      - name: get-action 
      - name: action
      - name: wait
        properties:
        - time: "10m"
      - name: evaluation
```

Based on the above example, the sequence waits 10 minutes before triggering the `evaluation`. 

❔ Implementing a separate service for just calling the wait function is maybe over-engineered. Can we integrate the functionality into an existing service OR should we model it as an extra property that is controlled by the shipyard controller? 

```
- name: "production" 
  sequences: 
  - name: remediation
      triggeredOn: 
      - event: production.problem.opened
      tasks:
      - name: get-action 
      - name: action
        properties: 
          - waitTime: "10m"
      - name: evaluation
```

**Task:**
- Discussion: Which is the best option to implement the wait functionality: a separate service or controlled by the shipyard-controller? 
- Implement the chosen approach. 

**Definition of Done:**
- [ ] When the shipyard contains a `wait` task configured by the `time` property, the sequence

johannes-b:
## keptn/keptn: Clean-up remediation-service to not control the remediation sequence

**Current status:** In its current implementation, the remediation-service controls the execution of the remediation sequence by sending out the `action.triggered` event and waiting for `action.finished` and `evaluation.finished`. Besides, waiting for 10 minutes is hard-coded into the services. 

**Clean-up remedidation-service:** The remediation-service does not control the sequence, but:
* it subscribes to a `get-action.triggered` event,
* it analysis the event payload and determines the next action to execute by looking into the `remediation.yaml`.
* If the `remediation.yaml` defines multiple actions for one problem and the one was already be executed, the remediation-service must be smart enough to take the next one in the list of actions. 

As a result of this clean-up task, the shipyard-controller takes over the responsibility for sending `action.triggered`, and `evaluation.triggered` - since this is modeled in the shipyard. 

❔ (1) Should the [sh.keptn.event.problem.open](https://github.com/keptn/spec/blob/master/cloudevents.md#problem) event work as trigger for the `remediation` sequence (2) how can we resolve: https://github.com/keptn/spec/issues/59

**Definition of Done**: 
- [ ] The remediation sequence execution is explicitly modelled by the shipyard, e.g.:
```
- name: "production" 
  sequences: 
  - name: remediation
      triggeredOn: 
      - event: production.problem.open
      tasks:
      - name: get-action 
      - name: action
      - name: evaluation
```

johannes-b:
## keptn/keptn Implement looping-mechanism via shipyard-controller

**Use-case:** A remediation config can contains multiple actions that wait to get executed as response to an open problem, e.g.:
`remediation.yaml`: 

```
apiVersion: spec.keptn.sh/0.1.4
kind: Remediation
metadata:
  name: carts-remediation
spec:
  remediations:
    - problemType: Response time degradation
      actionsOnOpen:
        - action: scaling
          name: scaling
          description: Scale up
          value: "1"
        - action: toggle-feature
          name: Toogle feature flag
          description: Toogle feature flag EnablePromotion to OFF
          value: 
            EnablePromotion: "off"
```

A remediation sequence will execute just one action: 

```
- name: "production" 
  sequences: 
  - name: remediation
      triggeredOn: 
      - event: production.problem.open
      tasks:
      - name: get-action                           < get action - "scaling"
      - name: action                                 < execute action - "scaling"
      - name: evaluation
```

To achieve a looping mechanism, the `remediation` sequence should get re-triggered when the `evaluation` task **failed**.

**Proposal:**
- Advance the trigger filters to consider the result of an individual task: 

```
- name: "production" 
  sequences: 
  - name: remediation
      triggeredOn: 
      - event: production.problem.opened
      - event: production.remediation.finished
          selector:
            match:
              evaluation.result: fail                             < see here
      tasks:
      - name: get-action 
      - name: action
      - name: evaluation 
```

- ❔ Is there a better way to define the *advanced* task filter?

**Tasks:**
- Update of shipyard in keptn/spec: https://github.com/keptn/spec/blob/master/shipyard.md
- Extend the shipyard-controller to filter events on task level

**Definition of Done:**
- [ ] When I have multiple actions in a `remediation.yaml` configured, they get executed by a single remediation sequence (get-action, action, evaluation) run.

johannes-b:
## keptn/examples  Prepare an SLO/SLI file that checks whether the Problem has been closed (as part of the remediation process)

*When using Dynatrace for problem detection:* 

The dynatrace-sli-service sends the number of open problems when fetching SLI values in course of a remediation process. The number of open problems should be taken into consideration when doing the evaluations task. Therefore, adapt the SLO (`slo-self-healing.yaml`) to contain a check for the open problems: 
```
  - sli: "problem_open"
    displayName: "Problem open"
    key_sli: true
    pass:             
      - criteria:
          - "=0"  
    weight: 1
```

Since the prometheus-sli-services does not return the number of open problems, a separate `slo-self-healing.yaml` file must be provided for this example. 

**Task:**
- Change the `slo-self-healing.yaml` for Dynatrace
- Adapt the tutorials accordingly

**Definition of Done:**
- [ ] When doing the self-healing tutorial with Dynatrace, the number of open problems is considered as part of the evaluation step

johannes-b:
## keptn-contrib/dynatrace-service Adapt the dynatrace-service to consider the ProblemDetailsJSON property containing the problem root cause

**Status quo:** Currently, the value of the *ProblemTitle* is used for resolving the problem<>acion mapping in the remediation.yaml config.

 **Improvement:** Since Dynatrace already knows the root cause of a problem, the root cause should be used for resolving the problem<>acion mapping. Unfortunatelly, the root cause is not part of the problem event sent out by Dynatrace, but can be derived: 
 
*How to get the root cause:*
- Use the `/problems/{problemId}` endpoint from the Dynatrace API while the `problemid` is contained in the *ProblemDetails* see `id`: 
```
      "ProblemDetails": {
        "displayName": "P-210325",
        "endTime": -1,
        "hasRootCause": true,
        "id": "-7814562899452185567_1617086820000V2",
        "impactLevel": "SERVICE",
        "severityLevel": "ERROR",
        "startTime": 1617087120000,
        "status": "OPEN"
      },
```
- The response contains the `evidence` property which contains the root cause of that problem: 

```
  "evidenceDetails": {
    "totalCount": 1,
    "details": [
      {
        "evidenceType": "EVENT",
        "displayName": "Failure rate increase",
        "entity": {
          "entityId": {
            "id": "SERVICE-1016B63497E3FEF6",
            "type": "SERVICE"
          },
          "name": "sockshop.carts.production"
        },
        "groupingEntity": null,
        "rootCauseRelevant": true,
        "eventId": "-7814562899452185567_1617086820000",
        "eventType": "FAILURE_RATE_INCREASED",
        "startTime": 1617086820000
      }
    ]
  }
```

- Add `rootCause`:`<value>` to the `problem` data block, while <value> = `eventType` 

**Tasks:**
- Adapt the remediation service to use the `rootCause` instead of the `problemTitle` (fallback still to `problemTitle`) for getting the proper remediation action
- Update the spec to support the new property `rootCause`
- Update of example files (`remediation.yaml`) to reference the value of the `rootCause` instead of `problemTitle`

**Definition of Done:**
- [ ] As a user, I can define the root cause in the `remediation.yaml` which will be mapped to the remedation action.

johannes-b:
## keptn/examples Update of tutorials/examples/docs/spec to support the remediation process modeled via Shipyard v.0.2

johannes-b:
## keptn/keptn: Clean-up of Bridge due to changed remediation use-case

https://github.com/keptn/keptn/issues/3698

jetzlstorfer (approved):
lgtm

PR #52, opened 2021-06-21
KEP 52 - Subscription management for Integrations

johannes-b:
# Subscription management for Integrations

**Mockup:** https://github.com/keptn/keptn/issues/1280

**Success Criteria:** Easy integration of Keptn-services into a Shipyard process. 

*Note:* The term *Keptn-service* is used as a synonym to an integration that is maintained in the https://github.com/keptn-contrib organization. A *Keptn-service* (aka. Integration)  has a subscription to an event and performs a task based on the event type and payload. 

## Motivation

* *Pain*: After implementing [KEP45](https://github.com/keptn/enhancement-proposals/pull/45), an integration is listed in the Uniform screen, and subscriptions are displayed. To change the subscription, a user needs to manually adapt the manifest of the deployed Kubernetes services (of this integration). This is not possible when no access to the Kubernetes Cluster is given.

* *Target*: Allow customization of subscriptions to integrate custom Keptn-services (Integrations) into a process definition using the Bridge.

* *Driver*: To foster the usage of custom integrations in Keptn, there must be UI/UX improvements in handling subscriptions of integrations.  

### As a user, I want to add multiple subscriptions to my integration by using the Bridge and then manually applying the change on a remote execution plane. 

A subscription has: 
* `name`: Name of the subscription for display purposes
* `event`: Event the subscription works on (e.g.: `sh.keptn.event.deployment.triggered`, `sh.keptn.event.release.triggered`)
* `filters`: On the project, stage, and service level

image

*User flow:*
- The button `Add subscription` opens an empty form to set `name` and to specify `event` and `filters`. 
image

### As a user, I want to add multiple subscriptions to my integration by using the Bridge and then automatically apply the changes on the control-plane.

*User flow:*
* See the first steps above. 
* If service is running on the control-plane, an `Update subscriptions` button is provided that updates the subscriptions on the cluster when clicking it. 
image

 ### As a user, I want to add parameters to my subscription to enrich it with context information. 

* A subscription supports a list of parameters; a parameter is a `key:value` pair. 

*open Questions:*
- The parameters won't be added by the shipyard-controller to the `.triggered` event because this is custom information and Keptn core should not expose information of other subscriptions. Hence, can the distributor of the integration take care of adding the parameters to the event payload before forwarding it to the receiving service? 

image

### As a user, I can validate whether my Shipyard (for this project) is covered by the uniform. If a missing subscription is found, I can add it with two clicks.

image

*User flow:*
[...]

PR #54, opened 2021-07-14
KEP 54 - Seamless integration of Webhook-based services

johannes-b:
# Seamless integration of Webhook-based services

**Success Criteria:** Seamless integration of Webhook-based services

## Motivation

* *Pain*: Writing/deploying custom integrations for just calling a webhook is cumbersome

* *Target*: As a user, I can specify a webhook that gets called for any task the `webhook-service` is acting on.

* *Driver*: Drive adoption with an easy and flexible way to integrate with other tools

*Note (a prerequisite):* See [KEP 52](https://github.com/keptn/enhancement-proposals/pull/52), which allows adding a new subscription for an integration. In this case, we are dealing with the `webhook-service`. Hence, the `webhook-service` is selected and a Keptn user adds a new subscription for this service: 

image

## General overview (and big picture)

image

Link to the Diagram: <>

### As a user, I can enter the configuration for the webhook call and the configuration is stored in the Git repo. The stage and service filters define where to store the config in the repo of this project. 

image

Next to the task (e.g., evaluation.triggered) and filter (stage/service) the webhook-service should act on, the webhook-service needs the following properties: 
- `request method`: [`POST`, `PUT`]
- `request URL`: webhook endpoint that is called
- list of `custom headers`: that are passed over to the endpoint (e.g., for authentication)
- `payload`
- `proxy` (optional)

The configuration (`webhook-conf.yaml`) of the webhook-service looks as follows:

```
apiVersion: webhookconfig.keptn.sh/v1alpha1
kind: WebhookConfig
metadata:
  name: webhook-configuration
spec:
  webhooks:
  - type: "sh.keptn.event.test.triggered"
    requests:
    - "curl --header "Content-Type: application/json" --request POST –data '{"event":"{{.event.type}}","service":"{{.event.data.service}}","environment":"{{.event.data.stage}}' https://api.mycompany.com"
```
After a user entered the information and clicked the [Save] button, a `webhook-conf.yaml` is stored in the Git repository of this project. The filter of the subscription defines the exact location where the `webhook-conf.yaml` is stored. 
* If no filter is selected, the config is stored on project level: `​/project​/{projectName}​/resource` 
* If a stage filter is selected, it is stored on stage-level:  `/project/{projectName}/stage/{stageName}/resource` 
* If a service is selected, it is stored on service-level: `/project/{projectName}/service/{serviceName}/resource`
* If a stage & service are selected, it is stored on service-level:  `/project/{projectName}/stage/{stageName}/service/{serviceName}/resource`

When updating this configuration and changing the filter (e.g, by adding a service), it is deleted first and then added again depending on the new location.

### As a user, I would like to use a secret value for the webhook call.

image

Almost every API is secured using a `x-token` or any other kind of API token to authenticate the request. This sensitive information cannot be stored as part of the webhook configuration (`webhook-conf.yaml`) since it would be accessible on the GitHub repository. Instead, it must be kept in a secret. This secret needs to be created first using the `Secrets` screen. Afterwards, the user proceeds as follows: 
1. Add an additional custom header - key-value pair - by clicking the **+** button, next to *Custom Headers* 
1. Change the type of the value from plain text to a secret-value. Click on the **lock** icon
1. Select the secret from where the value should be derived
1. Specify the secret-key that refers to the value in the secret

The webhook-service will take this information to get the proper secret value which is then included in the API call. 

The resulting webhook-conf.yaml looks as follows:

```
apiVersion: webhookconfig.keptn.sh/v1alpha1
kind: WebhookConfig
metadata:
  name: webhook-configuration
spec:
  webhooks:
  - type: "sh.keptn.event.test.triggered"
    envFrom:
    - secretRef:
      name: mysecret
    requests:
    - "curl --header "Content-Type: application/json" --header "x-token: ${SECRET_KEY}" --request POST –data '{"event":"{{.event.type}}","service":"{{.event.data.service}}","environment":"{{.event.data.stage}}' https://api.mycompany.com"
```

### As a user, I would like to define a custom payload that is forwarded to the webhook URL 

The `payload` field allows adding any kind of JSON payload which will be sent to the webhook. To map data from the incoming Keptn event to the payload, a templating mechanism is available.  

* A Keptn user can reference a property from the received event by using `"{{.event.[property]}}". The webhook-service will then replace this part with the actual value before sending it to the endpoint. 
* Example, of custom payload:
```
{ 
   "eventType":"{{.event.type}}",
   "service":"{{.event.data.service}}",
   "environment":"{{.event.data.stage}}",
   "deploymentUrl":"{{.event.data.deployment.deploymentUrl}}"
}
```
* A link in the Bridge should be provided that leads the user to an example of an event. 

### The webhook-service

* The webhook-service is a default Keptn-service
* It lives in: https://github.com/keptn-contrib organization
* It is integrated into the test suite of Keptn > tested with each new Keptn release

The functionality is as follows:
* When the webhook-service gets triggered (based on the event it is subscribed for), it first sends a `sh.keptn.event.[task].started` event
* Then, it derives the `webhook-conf.yaml` that is stored on service-level. 
  * If there is no configuration, the service looks on stage-level.
    * If there is no configuration, the service looks on project-level. 
* Translates the event payload of the incoming event into the custom payload for the webhook 
* Derives the secret value from the `secret` using the K8s API. ❓ Should we open the `/secrets` endpoint to GET secret values? 
* Sends the API call
* The webhook-service will set `result` of the `.finished` event to `pass` if the HTTP Webhook returned 2xx (200-299); if not, it will return `fail`.
* The response body of the Webhook call will be added to the `data.[task]` section of the `.finished` event
* Sends a `sh.keptn.event.[task].finished` event

### Security concerns

- Network Policy must decline cluster-internal calls

### DIFF
--- a/text/0054-webhook-service.md
+++ b/text/0054-webhook-service.md
@@ -1,0 +1,13 @@
+# Seamless integration of Webhook-based services
+
+**Success Criteria:** 
+
+## Motivation
+
+* *Pain*:
+
+* *Target*:
+
+* *Driver*:
+
+## User Cases: 


johannes-b:
### Issues:

- https://github.com/keptn/keptn/issues/3822

johannes-b:
# Feature requests for Bridge

### As a user, I would like to test my Webhook configuration before saving it

* After a successful test, the `Save` button gets activated


### As a user, I would like to see a previous event of the selected type

* The payload of the last event of the selected type is shown:

image

johannes-b:
### Enhancement

As mentioned in a comment on an implementation issue: https://github.com/keptn/keptn/pull/5152#issuecomment-914135380, the current approach does not allow to have multiple subscriptions with the same or overlapping filter configuration. For example: 

- Subscription 1: subscribes to `sh.keptn.event.test.triggered` for stage `dev`, `hardening`  --> here the user wants to send a slack notification
- Subscription 2: subscribes to `sh.keptn.event.test.triggered` for stage `dev` --> here the user wants to trigger a test tool

**Problem:**  If the webhook.yaml was created by another subscription (Subscription 1), but a new subscription (Subscription 2) has such an overlapping filter configuration, the Bridge will override the old configuration file without warning since it is not possible to differentiate between both (both are subscribed to `sh.keptn.event.test.triggered`, which is used as a unique identifier)

oleg-nenashev:
@johannes-b could you please summarize the current status of this KEP? It was not on the roadmap, and the right now it is hard to say what is left

johannes-b:
@oleg-nenashev, please find the remaining tickets for this KEP using this filter: https://github.com/keptn/keptn/issues?q=is%3Aissue+is%3Aopen+label%3Akep%3A54

I categorized these issues as follows:

### Know limitation

- https://github.com/keptn/keptn/issues/7042

### Migrate from `v1alpha1` to `v1beta1`

- https://github.com/keptn/keptn/issues/7322
- https://github.com/keptn/keptn/issues/7839
- https://github.com/keptn/keptn/issues/8128

### Testing

- https://github.com/keptn/keptn/issues/8027

### UX improvements - nice to have, the list is prioritized (top-down)

- https://github.com/keptn/keptn/issues/6892
- https://github.com/keptn/keptn/issues/6055
- https://github.com/keptn/keptn/issues/5952
- https://github.com/keptn/keptn/issues/6794

thisthat:
The feature has been implemented.

Issue #14, opened 2020-03-24
Allow external tools to provide SLI results via events vs just via SLI-provider


grabnerandi:
# Allow external tools to provide SLI Results via Events vs just via SLI Provider

SLI results should not only be pulled in via an SLI Provider but it must be possible to push SLI Results to Keptn from tools as they are taking part of the delivery process, e.g: JMeter should be able to provide SLI Results after test execution is done via the Test Finished Event as this is the time when these values are actually available. This also eliminates writing an SLI provider to fetch data that is already available

## Motivation

When executing JMeter tests, JMeter has a lot of valuable SLI data (response time, failure rate, ...) that should be made available to the Lighthouse service to be included in SLO validation. Letting a Keptn Service or even an external tool that sends e.g: a Deployment Finished Event provide SLI Results from this tool eliminates the need to implement custom SLI providers for each potential external data source. Instead - we giving those tools the chance to provide SLI results at the time when these tools have the data available.
This eliminates the need to write specific SLI Providers for tools that are already integrated with Keptn such as the JMeter or Neoload service. This option also allows tools that integrate with Keptn, e.g: GitLabs, Jenkins ... to send key metrics via e.g: the Keptn Start Evaluation Event. Examples here would be code coverage metrics, build times, ...

While we will still need support for Multi SLI Providers I think this approach will reduce the need for multi SLI providers by a great deal as most use cases I am aware of right now are data sources such as Testing Tools. The primare SLI Provider would then be - just as it is now - the Monitoring TOol SLI PRovider

## Explanation

I propose to allow any tool to send SLIResult data as part of a cloud event. It can be the same format as our internal IndicatorValues []*SLIResult that is part of the InternalGetSLIDoneEventData

Here are a couple of scenarios
#1: Testing Tool sends metrics as part of Test Finished
JMeter can send SLIResults as part of Test Finished which are then automatically available to the Lighthouse Service, e.g: jmeter_rt_test1, jmeter_rt_test2, ...
This allows our users to use these SLIs in their SLOs without having to specify them in a custom SLI. This brings immediate value for quality gates without having to configure an SLI provider and it allows us to reuse metrics that are already available by the test execution

#2: Pipeline sends SLIs as part of the Start Evaluation Event
Jenkins can for instance send SLIResults when it starts an evaluation event, e.g: jenkins_codecoverage, jenkins_buildtime, ... 
[...]

PR #56, opened 2021-07-21
KEP 56 - Resolving debt of SLI Retrieval

johannes-b:
# Resolving Debt of SLI retrieval (> enabling Multi-SLI Provider Support)

## Short abstract
_Extend the Keptn quality gate capabilities by using the event handling functionality of Keptn (and not the lighthouse-service)._

### Target audience pain points

As an *SRE/DevOps engineer* using Keptn for implementing a quality gate in a delivery workflow:
* I would like to query SLIs from multiple SLI providers (e.g., Dynatrace, Prometheus, Splunk, etc.) which are then used in a quality gate evaluation.
* I would like to take the SLIs from a test execution, without querying a monitoring tool. For example, a Jmeter test already provides metrics about the response time. This information should be consumed by the quality gate evaluation without fetching the data from another tool, but rather taking it from the `evaluation` event payload.
* I would like to query data through a service that is running on an execution plane. (Currently, an SLI-provider must run on the Keptn control plane).   
* I would like to subscribe a custom SLI-provider to a `get-sli` event using the Keptn subscription mechanism. (Currently, the `keptn configure monitoring` command is used for this step.)
 
## Investment / Use-case drivers

* Enriching the Keptn *quality gate use-cases* by supporting the above-mentioned requirements.

_Open discussion_
* https://github.com/keptn/keptn/discussions/4672

### What it is? What it is not?

* Resolving the design debt in the *lighthouse-service* that handles `get-sli` events. Due to this built-in functionality, limitations for certain use-cases (multi sli-provider support, running SLI-provider on execution-plane) exists. This is also the reason, why the `get-sli` task comes after `evaluation` what should be the other way around: 

image

* Deprecation of the `sh.keptn.event.monitoring.configure` event and the `keptn configure monitoring` command since the Uniform (event subscription) will manage which SLI-provider should react on a `get-sli` event. 

## Success criteria

Data from multiple SLI providers, as well as a testing tool that is part of a sequence execution, can be used for a quality gate evaluation. Additionally, an SLI provider can run on an execution plane. 

--- 

### Remarks
- This KEP uses the term **SLI-provider**. An SLI-provider is basically a typical Keptn-service that just listens on `get-sli` events. 
- ⚠️ This KEP uses the task name **get-sli**. However, this does not follow the naming schema of Keptn, as mentioned here: https://github.com/keptn/spec/issues/57 - Consider changing the name of the task as part of this KEP.

---

## Technical Implementation Details 

**Status quo:**
This section explains the current implementation of fetching SLIs from an SLI-provider. 

* Currently, the Shipyard looks as shown below, while the `evaluation` task encapsulates two types of events: (1) `get-sli.triggerd`, (2) `evaluation.triggered`
  ```
  apiVersion: "spec.keptn.sh/0.2.0"
  kind: "Shipyard"
  metadata:
    name: "shipyard-sockshop"
  spec:
    stages:
      - name: "dev"
        sequences:
          - name: "delivery"
            tasks:
              - name: "deployment"
              - name: "test"
              - name: "evaluation"
              - name: "release"
  ``` 
* The event handling for the `sh.keptn.event.get-sli.triggered` and `sh.keptn.event.get-sli.finished` events is managed by the *lighthouse-service*; which gets triggered by an `evaluation.triggered`
  * for `get-sli.triggered`: see: https://github.com/keptn/keptn/blob/master/lighthouse-service/event_handler/start_evaluation_handler.go#L56
  * for `get-sli.finished`: see https://github.com/keptn/keptn/blob/master/lighthouse-service/event_handler/evaluate_sli_handler.go#L69
* Since the event handling is in the *lighthouse-service*, these events are not exposed on the `​/event​/triggered​/{eventType}` endpoint of the *shipyard-controller*. Consequently, the SLI-provider has to run inside the Keptn control-plane for getting this type of event; running the SLI-provider on the execution-plane is not possible. 
* The *lighthouse-service* listens on a `sh.keptn.event.monitoring.configure` event to get the information which SLI-provider should be used. 
* The *lighthouse-service* has a dependency on the K8s API because for sending out a `get-sli` event the service derives the configured SLI provider first; see: https://github.com/keptn/keptn/blob/master/lighthouse-service/event_handler/start_evaluation_handler.go#L96

**Goal:**

* The handling of the `get-sli` event has to move from the *lighthouse-service* to the *shipyard-controller* to reuse the eventing capabilities of the *shipyard-controller*; meaning that, e.g., it will also be exposed for SLI-providers on the execution-plane. 
* This concludes that the Shipyard looks as follows: 
  ```
  apiVersion: "spec.keptn.sh/0.2.0"
  kind: "Shipyard"
  metadata:
    name: "shipyard-sockshop"
  spec:
    stages:
      - name: "dev"
        sequences:
          - name: "delivery"
            tasks:
              - name: "deployment"
              - name: "test"
              - name: "get-sli"                  <- explicitly mention the `get-sli` task
              - name: "evaluation"
              - name: "release"

  ```

* Backward compatibility: For ensuring backward compatibility, the following Shipyard *should* be supported: 
  ```
  apiVersion: "spec.keptn.sh/0.2.0"
  kind: "Shipyard"
  metadata:
    name: "shipyard-sockshop"
  spec:
    stages:
      - name: "dev"
        sequences:
          - name: "delivery"
            tasks:
              - name: "deployment"
              - name: "test"
              - name: "evaluation"
              - name: "release"
  ```
  * If no explicit `get-sli` task is mentioned, a *smart default* triggers this event before the `evaluation` task is executed. 

* Deprecate the `sh.keptn.event.monitoring.configure` event and the `keptn configure monitoring` command since the Uniform (event subscription) will manage which SLI-provider should react on a `get-sli` event. 

## User Cases

By moving the logic of the SLI retrieval from the *lighthouse-service* to the*shipyard-controller*, the following use cases are possible: 

### As a Keptn user, I would like to include multiple SLI-providers in a quality gate evaluation. 

When the `get-sli` task is part of a task sequence, multiple SLI-providers can subscribe to this task. The *shipyard-controller* will then take care of waiting for the subscribed SLI-providers before continuing with the `evaluation` task; a capability the *shipyard-controller* is capable of as of today. 

### As a Keptn user, I would like to run my SLI-provider on an execution-plane.

When the `get-sli` task is part of a task sequence, the *shipyard-controller* will provide a `get-sli.triggered` event on the `/event​/triggered​/{eventType}` endpoint. As a result, SLI-providers on the execution-plane can pull this event to fetch SLI when they are supposed to do. 

### As a Keptn user, I would expect that a quality gate evaluation considers the SLIs provided by a prior executed test.

This use case relates to this proposal: https://github.com/keptn/enhancement-proposals/issues/14

* Whit the implementation of this KEP, a test tool can return SLI values by adding an object consisting of `metric`, `value`, and `success` to the `[event].data.get-sli.indicatorValues` array. The same way a typical SLI-provider would do.
* The l*lighthouse-service* will then consider those values for the evaluation too.

```
{
  "specversion": "1.0",
  "id": "c4d3a334-6cb9-4e8c-a372-7e0b45942f53",
  "source": "source-service",
  "type": "sh.keptn.event.test.finished",
  "datacontenttype": "application/json",
  "data": {
    "project": "sockshop",
    "stage": "dev",
    "service": "carts",
    "get-sli": {
      "start": "2019-10-20T07:57:27.152330783Z",
      "end": "2019-10-22T08:57:27.152330783Z",
      "indicatorValues": [
        {
          "metric": "response_time_p50",
          "value": 1011.0745528937252,
          "success": true
        }
      ]
    }
  },
  "shkeptncontext": "a3e5f16d-8888-4720-82c7-6995062905c1",
  "triggeredid": "3f9640b6-1d2a-4f11-95f5-23259f1d82d6"
}
```
## Parts of the payload need to be discussed

The payload of the `get-sli.triggered` event is specified here: https://github.com/keptn/spec/blob/master/cloudevents.md#get-sli-triggered 

```
  "get-sli": {
    "sliProvider": "dynatrace",
    "start": "2019-10-28T15:44:27.152330783Z",
    "end": "2019-10-28T15:54:27.152330783Z",
    "indicators": [
      "throughput",
      "error_rate",
      "request_latency_p95"
    ],
    "customFilters": [
      {
        "key": "dynatraceEntityName",
        "value": "HealthCheckController"
      },
      {
        "key": "tags",
        "value": "test-subject:true"
      }
    ]
  }
```

1. The payload refers to the SLI-provider that is configured: `sliProvider` 
   * No need to send this information anymore. The user has to define by the Uniform (event subscription) which SLI-provider should gather SLIs. 
   * Consequently, the `keptn configure monitoring [dynatrace | prometheus]` will become deprecated.
1. The payload contains the `start`/`end` time to evaluate.
   * The SLI-provider needs to implement this functionality, which is currently implemented in the lighthouse-service. 
1. The payload contains information which is defined in the SLO.yaml: `indicators` and `customFilters` 
    * The SLIs provided all indicators from the SLI.yaml. If the lighthouse-service is missing an SLI, the service stops working with the error message: `Did not get all SLIs to evaluate the SLO`.

### DIFF
--- a/text/0056-sli-retrieval.md
+++ b/text/0056-sli-retrieval.md
@@ -1,0 +1,1 @@
+# SLI Retrieval


jetzlstorfer:
I might have an additional use case:

### As a Keptn user I want to have an evaluation for a whole task where multiple tests are run
For example, if I run locust tests and litmus chaos tests that are both triggered by the same `test` task, the evaluation should not consider the start and end time of one of those tests, but should consider the first start and the last end time of all task executions of the `test` task. With this we can ensure that the evaluation will be done for the whole testing task.

agrimmer:
From a technical perspective, we should also remove the dependency on the `lighthouse-config` config map because this breaks the shipyard/uniform concept, i.e. the shipyard only controls the process and the uniform allows to configure the tooling.

johannes-b:
A related topic that can be addressed by this KEP is: https://github.com/keptn/enhancement-proposals/issues/14

odubajDT:
During an enhancement of the quality-gates integration tests (triggering evaluation) we came across a problem (or a deadlock) we do not handle. The test scenario is:
- create project, service, add resources etc
- configure SLI provider (just a silly one for test purposes) -> won't response on any event
- add slo.yaml file
- trigger evaluation

The aboved described scenario leads this sequence to be stuck forever. Reason: When triggering evaluation, first the evaluation.triggered event is sent by shippy, after this lighthouse-service sends evaluation.started event which leads to setting the sequence state from triggered to started and getting the SLI from the SLI provider, so get-sli.triggered is sent by lighthouse-service and afterwards (if no get-sli.started  event is sent by SLI provider -> sequence is stuck in started state. The shippy does not take care of sequences in state started  as it does when they are in state triggered (it changes their state to timedOut  after some time if he does not receives the appropiate .started event.

thisthat:
The TC decided to NOT support this KEP and favor the development of the Lifecycle Toolkit.

PR #64, opened 2021-09-22
KEP 64 - Revamping Keptn Events to achieve 1.0

johannes-b:
# Revamping Keptn Events

**Success Criteria:** Keptn Events 1.0

## Motivation

In this KEP, improvements of the [Keptn Event spec](https://github.com/keptn/spec/blob/master/cloudevents.md) are collected. Please feel free to your ideas below. 

### Remove stage name from event type of [Task sequence events](https://github.com/keptn/spec/blob/0.2.3/cloudevents.md#task-sequence-events)

**Pain:** Task sequence events (or just Sequence events) have the stage name as part of the event type. For example, a delivery sequence in staging is triggered via: `sh.keptn.event.staging.delivery.triggered`, which contains the stage name `staging` in the event type. Additionally, the event payload contains the stage information in the data block too: 

```
{
  "data": {
    "project": "sockshop",
    "service": "carts",
    "stage": "staging",
    "status": ""
  },
  "id": "0fbf9445-affa-401e-a3ea-ed2cdd9eb987",
  "source": "shipyard-controller",
  "specversion": "1.0",
  "time": "2021-09-20T08:55:40.446Z",
  "type": "sh.keptn.event.staging.delivery.triggered",
  "shkeptncontext": "253f3870-69e5-4ee4-9e9e-03f819bbb94a",
  "shkeptnspecversion": "0.2.3"
}
```

**Target:** Remove the stage name from the event type: `sh.keptn.event.delivery.triggered` and relay on the `stage` property in the data block of the event payload. 

**Technical details:** 

Currently, the shipyard allows linking two stages by referencing the event from the previous stage as follows: 

```
    - name: "dev"
      sequences:
        - name: "delivery"
          tasks:
            - name: "deployment"
              properties:
                deploymentstrategy: "direct"
            - name: "test"

    - name: "staging"
      sequences:
        - name: "delivery"
          triggeredOn:
            - event: "dev.delivery.finished"
          tasks:
            - name: "deployment"
```

Consequently, the event type shows where the event is coming from. Based on the above example: from the `dev` stage. To support this linking mechanism, Keptn already supports `match` properties that allow modeling that: 

```
    - name: "dev"
      sequences:
        - name: "delivery"
          tasks:
            - name: "deployment"
              properties:
                deploymentstrategy: "direct"
            - name: "test"

    - name: "staging"
      sequences:
        - name: "delivery"
          triggeredOn:
            - event: "delivery.finished"
              selector:
                match:
                  stage: "dev"
          tasks:
            - name: "deployment"
```

**Question:** 

❓ Is it necessary to introduce a differentiator between sequence event and task event?

### DIFF
--- a/text/0064-keptn-events.md
+++ b/text/0064-keptn-events.md
@@ -1,0 +1,16 @@
+# Keptn Events
+
+**Success Criteria:** 
+
+## Motivation
+
+**Target audience pain points**
+
+
+**Evidence**
+
+## What
+
+
+## Use Cases
+


robertjahn:
I really think that a name like `sequenceid` would be better than `shkeptncontext` because that matches what is is and that way too it does not have `keptn` in it.  And if we we are open to adjusting than that, then the name `triggeredbyid` might be a better self subscribing name than just `triggerid`

robertjahn:
Suggestion # 1 - rename the attribute `shkeptncontext` to `sequenceId`

robertjahn:
Suggestion # 2 - rename the attribute `shkeptnspecversion ` to `specVersion `

thisthat:
Keptn events are not stable since we released Keptn v1 LTS.

PR #81, opened 2022-04-21
KEP 81 - Switch from branch to folder-based Git usage

johannes-b:
# Switch from branch to folder-based Git usage

**Success Criteria**: *Without impact on a Keptn user, the migration to the folder-based approach of storing artifacts in Git is performed.*

## Short abstract

This is a technical prerequisite for implementing: [KEP-70](https://github.com/keptn/enhancement-proposals/pull/70). In the course of refining [KEP-70](https://github.com/keptn/enhancement-proposals/pull/70), it has been figured out that the current branch-based approach of storing artifacts in Git is a blocker. Research has shown that a switch to a folder-based structure solves this blocker and is an enabler for additional use-cases. 

```
------------          ------------          ------------
|          |          |          |          |          |
| >KEP-81< |--------->|  KEP-70  |--------->| TBC      |
|          |          |          |          |          |
------------          ------------          ------------
```

⚠️ **Changing behavior of Git usage for Keptn users**: This KEP is a drastic change in the behavior of using Git in Keptn, which requires proper documentation updates. 

## Why

### Research and comparison

To provide a foundation for this enhancement proposal, research has been conducted. Useful links to recent blog posts can be found here and a summary is provided below: 
* [Stop Using Branches for Deploying to Different GitOps Environments](https://codefresh.io/about-gitops/branches-gitops-environments/)
  * [Multiple environments(dev, stage, ..,. prod ) example](https://github.com/argoproj/argocd-example-apps/issues/57)
  * [Best practices for promotion between clusters](https://github.com/argoproj/argo-cd/discussions/5667)
* [How to Model Your Gitops Environments and Promote Releases between Them](https://codefresh.io/about-gitops/how-to-model-your-gitops-environments-and-promote-releases-between-them/)
* [From Code to Production with GitOps](https://cloud.redhat.com/blog/from-code-to-production-with-gitops)

In order to derive a decision here, @agrimmer, @thisthat, @markuslackner, and I (@johannes-b ) have created a comparison of the **branch-based** and **folder-based** approaches:

|                                                                                                                                        	| A) Branch-based structure                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           	| B) Folder-based structure                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 	|
|----------------------------------------------------------------------------------------------------------------------------------------	|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------	|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------	|
|                                                                                                                                        	| * "as designed" and used                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            	| * implemented in resource-service<br>* but not rolled-out                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 	|
| **Git usage:**                                                                                                                         	| git checkout/pull master  < ?<br><br>git checkout -b quality<br><br>< add files / folders ><br><br>git commit -s -m "Added new stage quality"<br><br>git push                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       	| git checkout master<br><br>mkdir quality<br><br>< add files / folders ><br><br>git commit -s -m "Added new stage quality"<br><br>git push                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 	|
| **Advantages / Disadvantages in general:**                                                                                             	| **Advantages of this approach in general:** <br>:heavy_check_mark: Security concern: It is much easier to secure individual branches in a Git repository instead of folders in a single branch<br><br>**Disadvantages of this approach in general:**<br>:x: Pull requests and merges between different branches are problematic > Promotion is never a simple Git merge<br>:x: People are tempted to include environment-specific code and create configuration drift.<br>:x: Large number of environments > maintenance of all environments (and their branches) gets quickly out of hand<br>:x: The branch-per-environment model goes against the existing Kubernetes ecosystem (e.g., Helm / Kustomize)** 	| **Advantages of this approach in general:**<br>:heavy_check_mark: The order of commits on the repo is irrelevant<br>:heavy_check_mark: By only copying files around, you only take exactly what you need and nothing else (no configuration drift) <br>:heavy_check_mark: No need to use Git cherry-picks or any other advanced git method to promote releases<br>:heavy_check_mark: Free to make any change from any environment to either an upstream or downstream environment (without any constraints about the correct “order” of environments) → to support hot-fix strategies<br>:heavy_check_mark: File diff operations to understand what is different between environments in all directions<br><br>**Disadvantages of this approach in general:** <br>:x: Security concern: Git means are needed to implement security (manual approval, PRs, CODEOWNER file, etc.)  	|
| **Kustomize**                                                                                                                          	|                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     	| Kustomize has a base/overlay concept based on the folder structure. Example:<br><br>├── base<br>│   \|── deployment.yaml<br>│   \|── kustomization.yaml <br>│   └── service.yaml <br>└── overlays     <br>├── dev     <br>│   ├── kustomization.yaml     <br>│   └── patch.yaml     <br>├── prod     <br>│   ├── kustomization.yaml     <br>│   └── patch.yaml     <br>└── staging         <br>     ├── kustomization.yaml         <br>     └── patch.yaml<br>                                                                                                                                                                                                                                                                                                                            	|
| **Helm**                                                                                                                               	|                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     	| Helm allows modeling this using a base Helm chart with different versions of a values.yaml - Example: <br><br>chart/<br>   [...chart files here..] <br>common/<br>   values-common.yml <br>variants/<br>   prod/<br>      values-prod.yml<br>   non-prod/<br>      values-non-prod.yml<br>   [...other variants…]  <br>envs/<br>      prod-eu/<br>            values-env-default.yaml<br>            values-replicas.yaml<br>            values-version.yaml<br>            values-settings.yaml<br>    [..other environments…]<br>                                                                                                                                                                                                                                                       	|
| **Advantages / Disadvantages in the context of Keptn:**                                                                                	| :x: Performance implications (multiple git pulls, i.e., one for each stage)<br>:x: Only one sequence at a time can be processed<br>:x: Impossible to retrigger old sequences, because we loose the history once stage is deleted<br>:heavy_check_mark: Well tested                                                                                                                                                                                                                                                                                                                                                                                                                                 	| :heavy_check_mark: Performance gain / no read lock (one git pull)<br> *  For write operations, a lock is still needed.<br>:heavy_check_mark: Complete history if the same stage is added/deleted multiple times<br>:heavy_check_mark: Git commit id = keptn context → is shared across the sequence execution (simplifies retriggering of a sequence)<br>:heavy_check_mark: Easier to go through the stages (e.g., get the resources of a service - used by the Bridge)                                                                                                                                                                                                                                                                                                                                                                               	|
| **Is there a blocker for this KEP?**                                                                                                   	| Which branch will be the origin of the new branch (is it master/main)?<br> :warning: Cannot be HEAD of main/master because then the stage would e.g. contain the shipyard.yaml. Besides, it can`t be any other branch because then we would reuse its git commit history.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                	|                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           	|
| **Helps us to get to a point where we can connect to any repo.<br> (i.e., could also be a source-code repository of an application)** 	| :x: No, since we modify the repository by adding/deleting branches and which branch should be used as the origin?                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       	| :heavy_check_mark: Yes, as the only required is a .keptn  folder with the sub-folders for the stages. Given read permissions on that folder, Keptn could modify it by simple commits. <br><br>Theoretically, we could then bind a repo not only to a project but also to services as we just need the upstream repo + branch assuming we find a .keptn folder.                                                                                                                                                                                                                                                                                                                                                                                                                                               	|
| **Helps us to get to a point where we don`t need a Git repo anymore. <br><br>(i.e., could make use of an S3 bucket)**                  	|                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     	| :heavy_check_mark: Yes, we could go even one step further and eliminate the dependency to Git because Keptn then just needs a URL to a bucket (e.g. S3)  where a .keptn folder is available. From there the configuration is consumed, download S3.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          	|
|                                                                                                                                        	|                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     	|                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           	|
| **Conclusion:**                                                                                                                        	| Adding/Removing stages is based on the current approach                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             	| Adding/Removing stages is based on the new approach<br>→ Implies that we need to migrate repositories                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     	| 


## What

### What it is?

* **Migration** of Git repository from branch-based to folder-based. The `.keptn` folder lives on the default branch, which is in many cases the main/master branch.
  * The branch structure of existing upstream repositories stays, but the user gets the info that s/he needs to clean it up.

* **Folder structure:** Changing the way of storing artifacts in Keptn from a branch-based to a folder-based approach. 
  * The root folder is `.keptn` which is located in the default branch of the Git repository.
  * Keptn has write permissions on that `.keptn` folder. The other folders and files of the repo could be protected with a CODEOWNER file. 
  * The `.keptn` folder has the following top-level structure: 
   ```
   ./keptn
     |-- pase
     |-- service
     |-- stage
   ```

* **API:** Full support of API endpoints from `configuration-service` / `resource-service`

* **Documentation / Tutorials:** Since this change is an impact on the way Keptn users are working with their Git repository, documentation and tutorials need to reflect this changed behavior.  

* **Disclaimer** in the Keptn folder and part of the `metadata.yaml` file: `This folder is managed by Keptn. Changes within this folder are at your risk.`

#### Use cases: 

##### ⭐ Use Case: As a user, I create a new project using the traditional way. 

```
keptn create project sockshop --shipyard=./shipyard.yaml --git-token=**** --git-remote-url=https://github.com/my-test-org/sockshop.git
```

This creates the `.keptn` folder in the default branch (main/master) on the upstream repository:
```
./keptn
  |-- project
  |-- service
  |-- stage
```

##### ⭐ Use Case: As a user, I have an upstream repository configured and will migrate it to the folder-based approach.


##### ⭐ Use Case: As a user, I still can add resources/configuration to Keptn.

```
keptn add-resource --project=PROJECT --stage=STAGE --service=SERVICE --resource=FILEPATH --resourceUri=FILEPATH [flags]
```

### What it is not?

* Adding and removing stages. This is covered by: [KEP-70](https://github.com/keptn/enhancement-proposals/pull/70)

* Defaults for missing config: 

##### ⭐ Use Case: As a user, I can create defaults which are used as fallback for missing configuration. 

For instance, the [dynatrace-service](https://github.com/keptn-contrib/dynatrace-service) makes use of that by allowing the user to store a `dynatrace.conf.yaml` on the project level, which works as default for missing stage-specific configs:

```
./keptn
  |-- project
     |-- dynatrace/dynatrace.conf.yaml
  |-- service
  |-- stage
     |-- dev
     |-- staging
          |-- dynatrace/dynatrace.conf.yaml
     |-- production
          |-- dynatrace/dynatrace.conf.yaml
```

If the dynatrace-service requests a `dynatrace.conf.yaml` for staging: 
* (1) Fetch config from staging --> found.
* ⚠️ What to do with the config in project? (Merge?)

If the dynatrace-service requests a `dynatrace.conf.yaml` for dev: 
* (1) Fetch config from dev stage --> not found.
* (2) Fetch config from project level --> found. 

This implies that it is possible to override configuration by providing a hierarchy for entities. In Keptn, this hierarchy is currently defined by the Shipyard, which states that a project contains stages and services: 
* project `1` ---- `n` stage
* project `1` ---- `n` service

Besides, the hierarchy is defined by: 
* `project`
  * `stage`
    * `service`

In a nutshell, the folder structure needs to be flat in the sense of having one root folder with sub-folders for the various Keptn entities (project, stage, service). Additionally, Keptn must become aware how those entities are related to each other. This is defined by the Shipyard.

## To be discussed

Given the folder structure of: 

```
./keptn
  |-- project
  |-- service
  |-- stage
```

How can we model the use cases: `carts` service runs with replicaSet=1 in dev and replicaSet=2 in staging? Or, `carts` service uses `functional.jmx` in dev, `performance.jmx` in staging. 

To support that use case, some sort of hierarchy needs to be modeled in the folders (see `./services` section)

```
./project/
./project/jmeter
./project/job

./stage/dev/job/config.yaml
./stages/hardening/helm/values.yaml

./service/carts/helm/values.yaml                     < Note: This is not covered by the current API. 
./service/carts/stage/dev/job/config.yaml
./service/carts/stage/hardening/helm/values.yaml
```

![image](https://user-images.githubusercontent.com/729071/169474249-4474d475-9626-4a9c-84a7-b0e6cd341f4b.png)


## Open questions

- [x] How does the folder structure look like?
  - See above
- [x] Which kind of config is stored? Just deployment/test config or Keptn config as well?
  - For now, we still support both types of config.


christian-kreuzberger-dtx:
I'm very much in favour of this approach:
image

It allows us to easily distinguish between stages and services.

johannes-b:
### Idea for future improvements
... this is out of the scope of this KEP but has been derived as a future improvement. 

Currently, the `resource-service` provides dedicated endpoints to manage resources/config for those entities. To make this more generic, a *label-based* approach on the API could make sense and allow future changes of the entity model (which is current project-stage-service). In other words, provide one endpoint that takes labels to map resources/config to the right entity. Example:

* The CLI command of:
```
keptn add-resource --project=PROJECT --stage=STAGE --service=MY-SERVICE --resource=FILEPATH --resourceUri=MY-FILE
``` 
* would use:
```
POST /resource
        + query param: service=MY-SERVICE
```
* which results in:
```
./keptn
  |-- project
  |-- services
     |-- MY-SERVICE
(+)     |-- MY-FILE
  |-- stages
```

As shown by this example, the flags --projects=PROJECT and --stage=STAGE are obsolete.

*Future (future):* Allow adding new entities to model specific organizational requirements. For example, an organization does continuous delivery using `environments` and `regions` as an additional grouping element in stages. This could result in entity hierarchy as follows: 
* `project`
  * `stage`
    * `environment` 
       * `region` 
         * `service`

with a folder structure like:

```
./keptn
  |-- project
  |-- environments
  |-- regions
  |-- services
  |-- stages
```

agrimmer:
I see the use case "Use Case: As a user, I still can create defaults which are used as fallback for missing configuration." as out of scope for this KEP. For example, the logic to first look for the dynatrace.conf.yaml on service-, then on stage-, and finally on project-level is implemented in the dynatrace-service and not the resource-service.
Any objections?

johannes-b:
@agrimmer, I agree that this use case should be out of scope. I moved it to the **What is it not** section. 

Due to the discussion on a very flat folder hierarchy (see: https://github.com/keptn/enhancement-proposals/pull/81#issuecomment-1132555621), my thoughts led me in the wrong direction why I mentioned this use-case. If we want to support this use case, a new enhancement proposal needs to be submitted.

agrimmer:
When implementing this KEP, I would propose to stick to the current API schema to prevent breaking changes for integrations. Although we know that the current API schema includes bad designs like containing the resource URI as path parameter.
Especially, when we are introducing new entities like regions, environments, etc., we need to re-work the API anyway.

Furthermore, I would accept that the API does not provide any mechanism to work with files stored in the folder ".keptn/service/my-service/my-resource-foder/values.yaml" because resources at the service-level w/o a stage are also unsupported in the current branch-based model.

Summarizing, I am proposing to switch to the folder-based model but w/o introducing any breaking change.

johannes-b:
I agree with you in both aspects: 
*  This KEP sticks to the current API implementation
*  This KEP does not add the functionality to manage service resources. The current API disabled the endpoints for that functionality anyways: 
image


----
I tried to visualize both aspects in the above image:

image

--> see the mapping of endpoints to folders (entities)
--> see the red note in the example

thisthat:
The TC decided to don't support this KEP since we want to focus on KLT.

PR #58, opened 2021-08-11
KEP 58 - Snapshot release

johannes-b:
# Snapshot releases

## Problem statement

**Problem 1:** Currently, Keptn is mainly designed for the deployment and promotion of - more or less independent - microservices without dependencies. This behavior might become a challenge when multiple services are handled and should be tested together. Using the current approach, each service has to be promoted individually and therefore, each stage might run with version combinations that are not tested together.

**Problem 2:** Keptn assumes that each sequence is running in the scope of a service. As a result, it is not possible to define tasks that run once per project/stage which leads to the problem that every task that is taken by a sequence has to be designed in a way that it does not interfere with other ones. If there would be such - *more global* - scoped tasks, it would be possible to, e.g. create infrastructure per project but also run global tests, which might be evaluated on a per-service basis.

## Success Criteria

In the future, it should be possible to deploy larger-scale applications/platforms using Keptn. Therefore, Keptn should be able to deal with a set of services, but also to take tasks on project/stage scope.

As a result, the following success criteria are defined:

* A set of services can be defined in Keptn
* Services can be added and removed to/from the set
* The current set of services in a stage can be promoted in one step to the next stage
* Tasks can be taken on the scope of the set instead of services

### General overview

PlatformDeployment (1)
Image source: <>

### (proposed) Entity model

image
Image source: 

* Snapshot - A set of Keptn services
* Service - A Keptn service
* Service Sequence Instance (Service-SI) - Execution of a sequence (e.g., delivery) for one service
* Snapshot Sequence Instance (Snapshot-SI) - Execution of a sequence (e.g., delivery) for an entire snapshot
* `service version` - value from CI build
* `snapshot version` - defined by Keptn

## Use Cases

Overview:
1. As a platform operator, I want to define a set of services to test and deploy them together in further stages
1. As a platform operator, I want to deploy all services together to ensure that they are operating together properly
1. As a developer, I want to ensure that my service is tested to ensure that it is running properly on the platform
1. As a platform operator, I want to run tests involving all of the services to ensure that the platform runs properly
1. As a developer/platform operator, I want to know which service causes problems related to the tests to be able to find out where problems are
1. As a platform operator, I’d like to promote the whole group to the next stage to ensure that the preceding tests are still valid

#### As a platform operator, I want to define a set of services to test and deploy them together in further stages

[...]

PR #82, opened 2022-04-22
KEP 82 - Version awareness in Keptn

johannes-b:
# Version awareness in Keptn

**Success Criteria**: **

## Short abstract

After implementing [KEP-81](https://github.com/keptn/enhancement-proposals/pull/81) and [KEP-70](https://github.com/keptn/enhancement-proposals/pull/70), this enhancement proposal of introducing service/config versioning in Keptn makes the switch to the folder-based approach sound and applicable. Please note that this has already been discussed in a broader scope in [KEP-58](https://github.com/keptn/enhancement-proposals/pull/58).

```
------------          ------------          ------------
|          |          |          |          |          |
|  KEP-81  |--------->|  KEP-70  |--------->| >KEP-82< |
|          |          |          |          |          |
------------          ------------          ------------
``` 

### TL;DR:
* Versioning of configuration is needed to allow a user to trigger a delivery without knowing the Git commit Id. 
* Uploading config is a bulk operation that takes a ZIP or URL to a `./keptn` folder. 

### Target audience / Pain points / Related discussions

**Target audience:**

* DevOps engineer, who has to set up a delivery process for and end-to-end delivery.

**Pain:** 

* *How does a Keptn user know which version (!= Git commit id) to deploy?*

  * *Why is this actually a problem?* Because it could be the case that a user has no access to the Git repo storing the configuration and hence the Git commit is unknown.

* Currently, Keptn derives the version information from the `image:tag`. Relying on this tag is a too opinionated approach because delivery must not always be for an image. 

**Related discussions/issues:**

* [Introduce version as additional meta-data for a service](https://github.com/keptn/keptn/issues/7420)

## What

Please consider the derived drawing that is be explained below: 

Versioning in Keptn drawio

* Starting in the top left corner, we see a CI pipeline that is building, testing artifacts, and finally updating configuration to deploy. While in Keptn 0.14 this configuration is uploaded using `keptn add-resource` or manually committed to the upstream repo, uploading it via a `/config` endpoint that knows a `service`, `version` and `config` parameter would make it much easier. 
* The `config` parameter supports two formats:
  1) A ZIP archive following a folder structure that stores the config files as explained below.
  2) A URL that points to a `./keptn` folder with the same structure. This `./keptn` folder could be part of another repository, e.g., a source code repository that is used for service development.
* The `resource-service` processes the API request and stores the configuration in Git. Basically, it extracts or downloads the config and pushes it to the Git repository. The folder structure could look as follows - but need to be clarified: 
    ```
    ./base
    ./services
      |- carts
      |- payment
    ./stages
      |- dev
      |- hardening
      |- production
    ```
* Besides, the `resource-service` tags the commit using the service name and version: `svc-version`. It is important to have the service name as part of the tag since the repo is managing multiple services that eventually have the same version at some point in time. However, with the service name it becomes a unique key. 
* (*Future - just an idea for now*): Another variant of the `resource-service` could use an S3 bucket as the backend with an implementation of the versioning approach for that kind of storage. This could be based on the file name. 
* Now, let's assume a Keptn user (or CI) wants to trigger a deployment for carts in `v0.2.0`. **How does the person know that this version is stored with the Git commit `#85a3f2c`**? A very challenging task or actually not possible when the repo is not accessible. The only meta-data that is known to the person is the version; i.e., `v0.2.0`. Consequently, the only information that can be provided to Keptn at this point of time. 
* When executing the delivery sequence, this version number `v0.2.0` could be mapped to the Git commit id `#85a3f2c`. At the end, this is a technical detail. 

### What it is?

* Introducing `version` property in keptn/spec
* (Configuration can be versioned)
* Bulk upload of configuration with version attached
* Allow triggering a sequence containing the version
  ```
  {
    "data": {
      "project": "sockshop",
      "service": "carts",
      "stage": "dev",
      "version": "0.13.3",
     [ ... ]
    }
  }
  ```
* Bridge uses `version` with fall back to `image:tag` for displaying version information
* Bridge uses `version` with fall back to `buildId` for displaying version information (https://keptn.sh/docs/0.14.x/bridge/#services-view) 

#### ⭐ Use Case: As a user, I can perform a bulk upload of configuration which has a version attached

image

#### ⭐ Use Case: As a user, I can trigger a sequence with a reference to the version

This trigger: 

image

will become the event: 

```
{
  "data": {
    "project": "sockshop",
    "service": "carts",
    "stage": "dev",
    "version": "0.13.3",
   [ ... ]
  }
}
```

#### ⭐ Improvement: Bridge uses `version` continuously through the different screens

image

### DIFF
--- a/text/0082-version-awareness.md
+++ b/text/0082-version-awareness.md
@@ -1,0 +1,30 @@
+# Version awareness in Keptn
+
+**Success Criteria**: **
+
+## Short abstract
+
+*tbc* 
+
+```
+------------          ------------          ------------
+|          |          |          |          |          |
+|  KEP-81  |--------->|  KEP-70  |--------->| >KEP-82< |
+|          |          |          |          |          |
+------------          ------------          ------------
+``` 
+
+## Why
+
+### Pain
+
+## What
+
+### What it is?
+
+#### ⭐ Use Case: As a user, I
+
+
+### What it is not?
+
+## Open questions
--- a/text/0082-version-awareness.md
+++ b/text/0082-version-awareness.md
@@ -4,7 +4,7 @@
 
 ## Short abstract
 
-*tbc* 
+After implementing [KEP-81](https://github.com/keptn/enhancement-proposals/pull/81) and [KEP-70](https://github.com/keptn/enhancement-proposals/pull/70), this enhancement proposal of introducing service/config versioning in Keptn makes the switch to the folder-based approach sound and applicable. Please note that this has already been discussed in a broader scope in [KEP-58](https://github.com/keptn/enhancement-proposals/pull/58).
 
 ```
 ------------          ------------          ------------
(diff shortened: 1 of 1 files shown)


thschue:
This issue is mainly obsolete with the Lifecycle Controller, and I am no more really convinced that it makes sense to add this functionality to Keptn itself. Therefore, I propose to close this enhancement proposal (won't do).

@johannes-b, @thisthat: What do you think about this?

johannes-b:
With the new approaches of the [Keptn lifecycle-toolkit ](https://github.com/keptn/lifecycle-toolkit), this KEP is not relevant anymore.

PR #86, opened 2022-09-19
KEP-86: Keptn LTS Release Line and Keptn 1.0

oleg-nenashev:
It is a common practice for the CNCF projects to offer a Long-Term Support (LTS) release line. Such a release line is important for enterprise users, and it could help Keptn adoption. The proposal is to introduce a new LTS baseline which would be based on the current Keptn version of the Keptn core and key services.

This KEP suggests the following:

* Introducing a first ever LTS release line in Keptn, versioned as Keptn 1.0
* List of components that would be managed as a part of the LTS
* Scope and timeline for the 1.0 release
* Release policy for versions beyond 1.0
* Changes in the release policy for the common release baseline

Browsable version: https://github.com/keptn/enhancement-proposals/blob/kep/keptn-lts/text/0086-keptn-lts.md

Initial Keptn LTS KEP draft: https://docs.google.com/document/d/1RdFegnZrxjWxJAem9auaeVQ5_mKl5wFlwd6MgF1ot0s/edit?usp=sharing

Signed-off-by: Oleg Nenashev <o.v.nenashev@gmail.com>

### DIFF
--- a/text/0086-keptn-lts.md
+++ b/text/0086-keptn-lts.md
@@ -1,0 +1,220 @@
+# Keptn LTS Release Line and Keptn 1.0
+
+
+## Motivation
+
+It is a common practice for the CNCF projects to offer a Long-Term Support (LTS) release line.
+Such a release line is important for enterprise users, and it could help Keptn adoption.
+
+
+## Explanation
+
+The proposal is to introduce a new LTS baseline
+which would be based on the current Keptn version of the Keptn core and key services.
+
+This KEP suggests the following:
+
+* Introducing a first ever LTS release line in Keptn, versioned as Keptn 1.0
+* List of components that would be managed as a part of the LTS
+* Scope and timeline for the 1.0 release
+* Release policy for versions beyond 1.0
+* Changes in the release policy for the common release baseline
+
+
+## Internal details
+
+
+### Scope of the LTS
+
+Keptn LTS scope would include not just the Keptn core components and CLI,
+but also key services popular among Keptn users and considered critical for the ecosystem.
+Also, Keptn specification would be pinned.
+
+
+#### Keptn Core components
+
+* Keptn core
+* Keptn Bridge
+* Keptn CLI
+* Installation and packaging scripts
+* Helm charts
+
+
+#### Keptn Specification
+
+Keptn specification would be pinned for the LTS. There will be no breaking changes in the specification within the LTS release, and minimum changes at all. The specification documentation may be extended over time for quality reasons.
+
+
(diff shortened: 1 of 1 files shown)


johannes-b:
Hi @oleg-nenashev, 
(cc @StackScribe)

the topic regarding tutorials is not clear to me. The proposal states: 

```
Tutorials 2.0
* Interactive tutorials. Continue Killercoda adoption
* Revamp the tutorials site, archive the existing tutorials
* Update tutorials for Keptn 1.0 LTS
* TBD - Prometheus, ArgoCD, [TBD for Dynatrace and Datadog]
```

Does this mean we provide tutorials for Keptn 1.0 on Killercoda or based on the current solution (as mentioned by the third bullet point)?

text/0086-keptn-lts.md
```diff
@@ -0,0 +1,220 @@
...
+
+
+### Documentation changes
+
+In the LTS we want to improve maintainability of documentation and
+to land pending stories in the documentation working group.
+
+
+* Adopt Documentation-as-Code in Keptn where reasonable
+    * Support for distributed documentation: multiple repositories, branches and versions.
```
StackScribe, line 169: We aren't going to be able to achieve this by the time LTS is released, right?  Is the plan to silently roll to this in the first few months after LTS is released?

text/0086-keptn-lts.md
```diff
@@ -0,0 +1,220 @@
...
+* Landing open pull requests with valuable functionality
+
+Notably Keptn GitOps will remain in preview, or KEP-60 RBAC will not be shipped in the first LTS release.
+The timeline for these releases is likely to be delayed by the community.
+
+
+### Documentation changes
+
+In the LTS we want to improve maintainability of documentation and
+to land pending stories in the documentation working group.
```
StackScribe, line 165: Could you explain exactly what this means?

StackScribe:
@johannes-b Re: tutorials...  My take is that, going forward, we want to encourage doing tutorials using kilrcoda but we will not rule out tutorials done in the old-style format.  But tutorials are also very expensive, both to do originally and to maintain.  So new tutorials should be justified -- why the tutorial format is more appropriate than straight documentation or a video (and videos also require maintenance).  I would also like each tutorial to have a maintainer who is committed to keeping the tutorial up-to-date.  As I said, just my opinion...

oleg-nenashev:
> Does this mean we provide tutorials for Keptn 1.0 on Killercoda or based on the current solution (as mentioned by the third bullet point)?

@johannes-b Killercoda only. We will be archiving the current tutorials

PR #89, opened 2023-02-08
KEP 87: Lifecycle Toolkit - Context Propagation for GitOps

thschue:

### DIFF
--- a/text/0087-klt-traceid-propagation.md
+++ b/text/0087-klt-traceid-propagation.md
@@ -1,0 +1,4 @@
+# KEP0087 - Lifecycle Toolkit Parent Trace Id in Keptn Apps 
+
+## STATUS: DRAFTING
+
--- a/text/0087-klt-traceid-propagation.md
+++ b/text/0087-klt-traceid-propagation.md
@@ -3,17 +3,17 @@
 **State: DRAFTING**
 
 ## Motivation
-When promoting apps to other instances, an end user would like to see the whole flow across the stages in a single trace. This KEP proposes the usage of a parent trace ID, to create links between traces across stages.
+When promoting apps to other instances, an end user would like to see the whole flow across the stages in a single trace. This KEP proposes the usage of the [W3C trace-id](https://www.w3.org/TR/trace-context/#trace-id) to create links between traces across stages.
 
 ## Current Design
 Currently, the Keptn Lifecycle Toolkit (Controller) starts a new trace whenever a new version/revision of a KeptnAppVersion is created and there is no way to pass over a TraceId to an Application. As a result, the Trace ends when an application is deployed and there is no Link to the next stage. 
 
 ## Proposed Design
-This KEP proposes to add a new field to the KeptnApp CRD called `parentTraceId`. This field is optional and can be used to pass over a TraceId to the KeptnApp. The main use case for this field is to pass over the TraceId of the previous stage to the next stage, when promoting an application and therefore keep an end to end view of the whole flow.
+This KEP proposes to add a new field to the KeptnApp CRD called `previousTraceId`. This field is optional and can be used to pass over a TraceId to the KeptnApp. The main use case for this field is to pass over the TraceId of the previous stage to the next stage when promoting an application and therefore keep an end to end view of the whole flow.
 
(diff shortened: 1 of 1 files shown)


text/0087-klt-traceid-propagation.md
```diff
@@ -0,0 +1,21 @@
+# KEP0087 - Lifecycle Toolkit Parent Trace ID in Keptn Apps 
+
+**State: DRAFTING**
+
+## Motivation
+When promoting apps to other instances, an end user would like to see the whole flow across the stages in a single trace. This KEP proposes the usage of a parent trace ID, to create links between traces across stages.
```
thisthat, line 6:
```suggestion
When promoting apps to other instances, an end user would like to see the whole flow across the stages in a single trace. This KEP proposes the usage of the [W3C trace-id](https://www.w3.org/TR/trace-context/#trace-id) to create links between traces across stages.
```

text/0087-klt-traceid-propagation.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Current Design
+Currently, the Keptn Lifecycle Toolkit (Controller) starts a new trace whenever a new version/revision of a KeptnAppVersion is created and there is no way to pass over a TraceId to an Application. As a result, the Trace ends when an application is deployed and there is no Link to the next stage. 
+
+## Proposed Design
+This KEP proposes to add a new field to the KeptnApp CRD called `parentTraceId`. This field is optional and can be used to pass over a TraceId to the KeptnApp. The main use case for this field is to pass over the TraceId of the previous stage to the next stage, when promoting an application and therefore keep an end to end view of the whole flow.
```
thisthat, line 12: I think `parent` is a bad name since it is not really a parent-child relationship. I would suggest using `previous`.

```suggestion
This KEP proposes to add a new field to the KeptnApp CRD called `previousTraceId`. This field is optional and can be used to pass over a TraceId to the KeptnApp. The main use case for this field is to pass over the TraceId of the previous stage to the next stage when promoting an application and therefore keep an end to end view of the whole flow.
```

malon875875:
- the single parentTraceId will therefore keep track of the multiple traces, is it correct to say that there will be one trace per stage?
- the promotion of applications will happen to different stages, not all stages are in the same k8 cluster
  - I don’t know OpenTelemetry enough in order to judge if a single parentTraceId can show the traces from k8-cluster-**A**.stage-dev -> k8-cluster-**B**.stage-test -> k8-cluster-**C**.stage-prod

malon875875:
+1
I would love to see this PR implemented within the [Keptn Roadmap](https://github.com/orgs/keptn/projects/10/views/5)

mowies:
@thschue do you see that field in the `spec`?
IMHO we should avoid a scenario where we mix data (i.e. KeptnApp) with metadata about it (i.e. the previous trace ID).
If we go with this, I would put the trace ID into the `metadata.labels` or `metadata.annotations`.
But I almost expect that we will pass along more metadata between stages in the future. Labels/annotations still don't sound future-proof/flexible enough to me. Why not go with a separate CRD that holds metadata instead?

thisthat:
> @thschue do you see that field in the `spec`? IMHO we should avoid a scenario where we mix data (i.e. KeptnApp) with metadata about it (i.e. the previous trace ID). If we go with this, I would put the trace ID into the `metadata.labels` or `metadata.annotations`. But I almost expect that we will pass along more metadata between stages in the future. Labels/annotations still don't sound future-proof/flexible enough to me. Why not go with a separate CRD that holds metadata instead?

Hey @mowies, the TC discussed your suggestion and we have some concerns regarding adding a new CRD. With a new CRD, we could get into sync issues if you use a GitOps tool. We decided that, for the moment, we go ahead and add the trace-id field into the `spec`.

bradmccoydev (approved):

Cyanopus:
@thschue Have you thought about promotion over different clusters- workloads across multiple clusters? You will have to have a synchronization piece, a mutex table of sort which stops multiple stages from syncing on pre-deployment tasks. Today promotion across different k8s clusters in different stages is a very important piece in the enterprise world.

PR #90, opened 2023-02-08
KEP 88: Lifecycle Toolkit - Adding Instances to KeptnApp

thschue:

### DIFF
--- a/text/0088-klt-instances.md
+++ b/text/0088-klt-instances.md
@@ -1,0 +1,3 @@
+# KEP 88: Adding Instances to KeptnApp 
+
+## STATUS: Drafting


text/0088-klt-instances.md
```diff
@@ -0,0 +1,102 @@
...
+
+Therefore, there are no explicit references to the instance of the application in the lifecycle toolkit resources.
+
+
+## Assumptions / Definitions
+- In the Keptn Lifecycle Toolkit, an instance describes a single instance of a workload or application. This can be a single stage or environment of a workload or application.
+- An instance has to be specified on the workload and application level
+- If an instance is not specified, the lifecycle toolkit will assume that there is only a single instance of the workload or application
+- If an instance is not specified on the workload level, the lifecycle toolkit will lookup the corresponding application and use the instance specified there
+- Properties of an instance are defined in the `KeptnApp` resource
```
thisthat, line 21: I would also add the propagation of this info into the span via OTel attribute
```suggestion
- Properties of an instance are defined in the `KeptnApp` resource
- Properties of an instance are stored as attributes in the generated Spans
```

text/0088-klt-instances.md
```diff
@@ -0,0 +1,102 @@
...
+      ( path: <path to the manifest> )
+      ( reposecret: <secret to access the repository> )
+    - name: staging
+      successor: 
+      - production
+    - name: production
+
+```
+
+The `instance` property is used to specify the current instance of the application. If this property is not specified, the lifecycle toolkit will assume that there is only a single instance of the application called "single".
```
thisthat, line 58:
```suggestion
The `instance` property is used to specify the application's current instance. If this property is not specified, the lifecycle toolkit will assume that there is only a single instance of the application called "single".
```

thisthat: I would suggest another name as default:

```suggestion
The `instance` property is used to specify the current instance of the application. If this property is not specified, the lifecycle toolkit will assume that there is only a single instance of the application called "main".
```

```suggestion
The `instance` property is used to specify the current instance of the application. If this property is not specified, the lifecycle toolkit will assume that there is only a single instance of the application called "default".
```

text/0088-klt-instances.md
```diff
@@ -0,0 +1,102 @@
...
+```yaml
+apiVersion: keptn.sh/v1alpha3
+kind: KeptnApp
+metadata:
+  name: podtatohead
+  namespace: keptn
+spec:
+  instance: <name of the current instance>
+  [...]
+  instances:
```
thisthat, line 37: This schema would require having the same `KeptnApp` with the instances structure across (potentially) multiple repositories. The benefit of this proposal is that we have a complete overview of the instances' tree.
I suggest only storing the partial view in each `KeptnApp`, i.e., the next successors of the current instance. This way, if we want to add a successor to, let's say, production, we don't need to change the `KeptnApp` of dev, staging, etc. 
As a drawback, this won't allow a complete overview in the `KeptnApp` definition and push this into the Observability platform via trace.

malon875875:
- An instance has to be specified on the workload and application level - agreed
  - Therefore KeptnApp will contain the information about the instance: dev - successor staging; staging - successor production….
- For the reposecret, I am thinking we could use [Sealed Secrets](https://medium.com/codex/sealed-secrets-for-kubernetes-722d643eb658)
- Can Keptn LCT manage multiple instances across clusters k8-cluster-**A**.stage-dev -> k8-cluster-**B**.stage-test ?

thschue:
> * An instance has to be specified on the workload and application level - agreed
>   
>   * Therefore KeptnApp will contain the information about the instance: dev - successor staging; staging - successor production….
> * For the reposecret, I am thinking we could use [Sealed Secrets](https://medium.com/codex/sealed-secrets-for-kubernetes-722d643eb658)
> * Can Keptn LCT manage multiple instances across clusters k8-cluster-**A**.stage-dev -> k8-cluster-**B**.stage-test ?

Hello and thank you for your comments!

* The KeptnApp would have all information needed to promote this through the stages
* Although it might not be the best idea to introduce a dependency to external/sealed secrets in keptn, I think this might be a good thing to document as soon as the feature is available :-) 
* The management of multiple instances depends on the tasks which are using the structure which is defined here. This EP introduces the CRD changes needed to support multi-instance operations. The implementation of the promotion itself will be only a KeptnTask ... In my mind, this is possible as the promotion should happen via git (at least in a GitOps setup) and this is more or less independent on where the clusters are

malon875875:
+1
I would love to see this PR implemented within the [Keptn Roadmap](https://github.com/orgs/keptn/projects/10/views/5)

text/0088-klt-instances.md
```diff
@@ -0,0 +1,102 @@
...
+    defaults:
+      repository: <git repository>
+      branch: <git branch>
+      path: <path to the manifest>
+      reposecret: <secret to access the repository>
```
thisthat, line 42: TC discussion: here we could use `secretName` and `secretKey` to fetch a K8s secret.

PR #92, opened 2023-02-09
KEP 90: Lifecycle Toolkit - Runtime Adaptions

thschue:

### DIFF
--- a/text/0090-klt-runtime-changes.md
+++ b/text/0090-klt-runtime-changes.md
@@ -1,6 +1,6 @@
 # KEP 90: Functions - Runtime Adaptions
 
-**State: DRAFTING**
+**State: APPROVED**
 
 ## Motivation
 Keptn Functions provide an easy way to add functionality to the delivery process. At the moment, the runtime is implemented in a very simple way, it is a simple container which executes deno functions. If the function succeeeds, everything is fine, if it fails, the container exits with an error code. This behavior has some drawbacks:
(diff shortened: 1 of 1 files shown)


agardnerIT:
Love this idea and support. 1 suggestion: For the returned JSON, can we make this [cloudevents ](https://cloudevents.io) compliant?

Can this event also either be a standard cdevent (it doesn't quite fit but perhaps a [testcase-finished](https://github.com/cdevents/spec/blob/v0.1.1/continuous-integration-pipeline-events.md#testcase-finished) OR if there isn't [a suitable event type](https://cdevents.dev/docs/) - we propose a new set of events to cdevents.

bradmccoydev:
@agardnerIT Test event types are being developed for cdEvents and there is a discussion on this at the moment, Ole from Kubeshop is leading this I will send a link to the draft when i find it. some context might also be in here: https://hackmd.io/2FRGlw9fTMmKN1OQUVvguA

thschue:
One option to get more insights into the runtime might be a running sidecar with an API exposed to the Controller (and secured by network policies), which allows the lifecycle controller to fetch the status, gather the results, or stop the job.

thisthat (approved):

AnaMMedina21 (dismissed):
Approved! bash and python would definitely be great additions for

> One option to get more insights into the runtime might be a running sidecar with an API exposed to the Controller (and secured by network policies), which allows the lifecycle controller to fetch the status, gather the results, or stop the job.

@thschue; Are you thinking about the sidecar helping Keptn infer the runtime specs without user input? I do think that fetching status/results would be a nice to have

bradmccoydev (dismissed):

thisthat (approved):

bradmccoydev (approved):

PR #96, opened 2023-02-10
KEP 91: Lifecycle Toolkit - Workload Information in KeptnApp

thschue:

### DIFF
--- a/text/0091-klt-workload-selectors.md
+++ b/text/0091-klt-workload-selectors.md
@@ -1,0 +1,56 @@
+# KEP 91: Lifecycle Toolkit - Workload Information in KeptnApp
+
+**State: DRAFTING**
+
+## Motivation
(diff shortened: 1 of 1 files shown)


malon875875:
This would greatly improve configuring Keptn LCT, so that not both Deployment manifest and KeptnApp Resource have to be modified any longer

nlamirault:
Another improvment could be to specify which namespace to target. Like that:

```yaml
apiVersion: keptn.sh/v1alpha1
kind: KeptnApp
metadata:
  name: my-app
  namespace: keptn
spec:
  version: 1.0.0
  workloads:
  - name: my-workload
    namespace: my-namespace
    version: 1.0.0
    selector:
      labelSelector:
        matchLabels:
          app: my-workload
          version: 1.0.0
    pre-deployment-tasks: my-pre-deployment-task (optional)
    post-deployment-tasks: my-post-deployment-task (optional)
    pre-deployment-evaluation: my-pre-deployment-evaluation (optional)
    post-deployment-evaluation: my-post-deployment-evaluation (optional)
```

AnaMMedina21:
With the comments on KEP92 [#97], could it just infer based on namespaces?

thisthat (changes requested):
Talking with @mowies, we believe this is not valid anymore since the `KeptnApp` is automagically created by the lifecycle operators watching annotations/labels on Workload level. Now users have to change only their K8s workload manifests.

PR #97, opened 2023-02-15
KEP 92: Lifecycle Toolkit - Separation of concerns for Tasks and MetricProviders

thschue:


AnaMMedina21 (commented):
I think the assumption of different namespaces is very realistic as we look at getting folks onboarded to KLT, especially for larger enterprise environments 

With this KEP, Keptn installation would be a global namespace by default and then allow folks to define it for each `KeptnTaskDefinition` or `KeptnMetricsProvider`, and if not defined it infers to the namespace Keptn was installed in?

thisthat:
> I think the assumption of different namespaces is very realistic as we look at getting folks onboarded to KLT, especially for larger enterprise environments
> 
> With this KEP, Keptn installation would be a global namespace by default and then allow folks to define it for each `KeptnTaskDefinition` or `KeptnMetricsProvider`, and if not defined it infers to the namespace Keptn was installed in?

Hey @AnaMMedina21, a deployment is usually namespaced. By default, KLT tries to look for Keptn CRDs in the same namespace where the current deployment occurs. However, instead of copy/paste the same CRD Task in multiple namespaces, it is better to "centralized" in the namespace where KLT is installed.

thisthat (dismissed):

text/0092-klt-concern-sep-tasks-providers.md
```diff
@@ -0,0 +1,29 @@
+# KEP 92: Separation of Concerns for Tasks and MetricProviders
+
+**State: DRAFTING**
```
thisthat, line 3:
```suggestion
**State: IMPLEMENTED**
```

thisthat (approved):

text/0092-klt-concern-sep-tasks-providers.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Current Design
+At the moment, it is assumed that KeptnTaskDefinitions as well as MetricProviders are deployed in the same namespace as the Application/Workload which gets deployed. Therefore, the Task Controller will look for a Metric Provider in the same namespace to compose the Task to be executed, the same is happening for the metrics with the metrics provider.
+
+## Assumptions / Definitions
+* A separation of concerns for tasks and metric providers is desired
+* If an object is defined in the same namespace as the Application/Workload, this is preferred over the global objects
+
+## Proposed Design
+The current design will be extended to support the separation of concerns. Therefore, the Task and Metrics Controller will try to find the respective objects in the same namespace as the Applications/Workloads and if this is the case, the Controllers will use this. If the object is not found in the same namespace, The Controllers will look for the object in the global namespace, where the Keptn installation is running.
```
mowies, line 18: the proposed design was implemented [here](https://github.com/keptn/lifecycle-toolkit/blob/5c14bf59e813db10f953ea019c8d61d7ec2e8f6d/lifecycle-operator/controllers/common/helperfunctions.go#L69-L83)

text/0092-klt-concern-sep-tasks-providers.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Possible Extensions
+* The Keptn installation can be configured to use a different namespace for the global objects
+* The TaskDefinitions/MetricProviders may be configured in a <namespace>/<object> way, to allow the user to specify the namespace where the object is located
```
mowies, line 22: I like those suggestions a lot, they would make a lot of sense for users I think. Just providing the option to use the Keptn namespace is probably not the ideal solution for users.

mowies (approved):

PR #105, opened 2023-08-06
KEP96 - Rename Tasks and Evaluations to Hooks

agardnerIT:
# KEP96

Proposes we (as a project) change the way we talk about KeptnTasks and KeptnEvaluations.

Rather than confusing users, this KEP proposes we discuss both as "hooks". There are thus two types of hook a task and an evaluation.

Note: This doesn't change the underlying implementation of CRDs. It's (probably) purely a documentation / messaging thing.

### DIFF
--- a/text/0096-rename-tasks-evaluations-hooks.md
+++ b/text/0096-rename-tasks-evaluations-hooks.md
@@ -1,0 +1,47 @@
+# Group Keptn Tasks and Evaluations as Hooks
+
+Proposal to change the way the project refers to Keptn Tasks and Evaluations. Instead, refer to both as Keptn Hooks.
+
+## Motivation
+
+* Make it simpler for a user to understand the core concepts of Keptn Lifecycle Toolkit
+
+## Explanation
+
+Hooks are a fairly standard and well understood software mechanism. So rather than referring to tasks and evaluations - then attempting to explain the difference.
+
+Instead we start referring to Keptn hooks. Keptn hooks are a way to interject into the lifecycle of a running Keptn process.
+
+With this description, the explanation becomes easy. There are two possible Keptn hooks:
+
+- Evaluations
+- Tasks
+
+Keptn hooks can be configured to fire at any of these times:
+
+- Pre-deployment at a workload level
+- Pre-deployment at an application level
+- Pre-deployment at a workload level
+- Post-deployment at a workload level
+
+## Internal details
+
+Note: This KEP does not mean changing anything about the current implementation: KeptnTasks and KeptnEvaluations remain as-is.
+
+This is simply a way to all agree on how we message and refer to these capabilities.
+
+Ergo, no additional coding is expected (apart from documentation rewording).
+
+## Trade-offs and mitigations
+
+
+## Breaking changes
+
+None
+
+## Prior art and alternatives
+
+
+## Open questions
+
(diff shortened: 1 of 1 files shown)


text/0096-rename-tasks-evaluations-hooks.md
```diff
@@ -0,0 +1,47 @@
...
+With this description, the explanation becomes easy. There are two possible Keptn hooks:
+
+- Evaluations
+- Tasks
+
+Keptn hooks can be configured to fire at any of these times:
+
+- Pre-deployment at a workload level
+- Pre-deployment at an application level
+- Pre-deployment at a workload level
```
StackScribe, line 24: Do you mean "Post-deployment at a workload level" here?

text/0096-rename-tasks-evaluations-hooks.md
```diff
@@ -0,0 +1,47 @@
...
+- Post-deployment at a workload level
```
StackScribe, line 25: I may be missing something but it seems like this requires changing the names of the annotations and sort of blurring the distinction between evaluations and tasks, at least at the configuration level.  Do we have evaluations and tasks that are executed only at the KeptnApp level?  I thought all the evaluations and tasks defined for the individual workloads were executed for any KeptnApp that includes those workloads.

I'm looking forward from comments from people with more knowledge!

text/0096-rename-tasks-evaluations-hooks.md
```diff
@@ -0,0 +1,47 @@
...
+
+Proposal to change the way the project refers to Keptn Tasks and Evaluations. Instead, refer to both as Keptn Hooks.
+
+## Motivation
+
+* Make it simpler for a user to understand the core concepts of Keptn Lifecycle Toolkit
+
+## Explanation
+
+Hooks are a fairly standard and well understood software mechanism. So rather than referring to tasks and evaluations - then attempting to explain the difference.
```
StackScribe, line 11: To me, the distinction between an evaluation and a task is fairly obvious.  What I think is confusing is the apparent inability to do an evaluation then trigger a particular task as a result of the evaluation, although this is a deliberate design decision.  To clarify this, we need a good, complex example of a KeptnApp that includes multiple workloads and uses ArgoCD to implement the "if-this-evaluation-result-then-do-this-logic" -- and I'm not sure whether either the KeptnEvaluation or KeptnTask will be part of that logic.

mowies (approved):
I like the idea! Although I think we should update this in the docs at least so that we get the word out as well. The code probably doesn't need to change at all

text/0096-rename-tasks-evaluations-hooks.md
agardnerIT: "To me, the distinction between an evaluation and a task is fairly obvious".

In a way, that's exactly my point in this PR. We (as those who know about Keptn) know about these terms. But if you are new to Keptn (and probably new or new-ish) to DORA / SRE / SLIs / SLOs etc. then a `Keptn Task` and a `Keptn Evaluation` will mean very little to you.

The term "hooks" is more generic, more widely used and thus more approachable. We don't remove the `Keptn Task` and `Keptn Evaluation` nomenclature - we just explain that both "Keptn tasks and Keptn evaluations" are types of hooks.

PR #107, opened 2024-02-01
KEP 98: Promotion Phase for GitOps use cases

mowies:


text/0098-promotion-phase.md
```diff
@@ -0,0 +1,42 @@
+# Promotion Phase for GitOps use cases
+
+Add a new lifecycle phase after post deployment dedicated to GitOps promotion tasks.
+
+**State: DRAFTING**
+
+## Motivation
+
+Since one of the main use cases for Keptn is enhancement of GitOps workflows, we should have bigger focus on that
+by introducing a dedicated stage after the post-deployment phase for GitOps promotion tasks.
```
thisthat, line 10: To emph that's the very last thing happening and not at workload level

```suggestion
by introducing a dedicated stage after the post Application deployment phase for GitOps promotion tasks.
```

text/0098-promotion-phase.md
```diff
@@ -0,0 +1,42 @@
...
+phase.
+
+## Prior art and alternatives
+
+The alternative to this KEP is that users re-purpose the post-deployment phase for promotion, which would limit
+flexibility since users lose the option to not promote when post-deployment tasks fail.
+
+## Open questions
+
+For now, none.
```
thisthat, line 38: how to deal the case where the App get deployed but a Task in the promotion step fails. will be the overall deployment reported as failure or success? 🤔

mowies: i think we should have a config option for users to define whether keptn will mark this as failed or not. similar to the non-blocking behaviour described [here](https://github.com/keptn/lifecycle-toolkit/issues/1934)

mowies: i added some content about this

mowies:
@thisthat pls re-review :)

text/0098-promotion-phase.md
```diff
@@ -0,0 +1,47 @@
+# Promotion Phase for GitOps use cases
+
+Add a new lifecycle phase after post deployment dedicated to GitOps promotion tasks.
+
+**State: DRAFTING**
+
+## Motivation
+
+Since one of the main use cases for Keptn is enhancement of GitOps workflows, we should have bigger focus on that
```
thisthat, line 9: I'd suggest dropping the GitOps wording in this section. The promotion task could potentially promote by applying the manifests directly into another cluster 😅

mowies: fixed

thschue:
This EP makes much sense and enables many workflows (especially regarding promotion tasks). Could it be that this partially deprecates this one: https://github.com/keptn/enhancement-proposals/pull/98 (on-success tasks but implemented in a new phase)?

mowies:
@thschue it's connected to the other KEP a bit, yes. But only in part of course. The thing is that you can't reference tasks here. So you basically have an unordered list of  tasks that are run as a result of the success of the post-deployment phase.
But there's no way of referencing specific tasks when a specific post-deployment task succeeded for example.

thschue (approved):
lgtm

thisthat (approved):

heckelmann:
Two questions: Will this phase being executed as well when the status of a post deployment task or evaluation is failing? If so, do I get the results in my promotion task as well?

mowies:
No, it will only execute if the phases before it succeeded.

mowies:
epic ticket for this: https://github.com/keptn/lifecycle-toolkit/issues/2907

